PR #18, opened 2011-10-05
array and object literals should be collapsable the way functions are

db48x:
Make js2-mode-toggle-element and js2-mode-hide-element handle array and object literals, plus a little refactoring so that they don't both do the same work.


devnill:
I would love this feature too!

PR #86, opened 2013-02-18
Consider NodeJS Globals

azer:
http://nodejs.org/api/globals.html

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -147,7 +147,12 @@
             DocumentRange Range RangeException
 
             ;; W3C XML
-            XPathResult XMLHttpRequest))
+            XPathResult XMLHttpRequest
+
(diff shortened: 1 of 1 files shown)


aguynamedben:
Is there a way to simply turn off js2-mode's globals checking? I want my inter (ESLint via flycheck) to handle things like this.

aguynamedben:
Answered my question with (using use-package)...

```elisp
(use-package js2-mode
  :ensure t
  :mode "\\.js\\'"
  :config
  (setq-default js2-mode-show-parse-errors nil)
  (setq-default js2-mode-show-strict-warnings nil))
```

Place in the code I found it: https://github.com/mooz/js2-mode/blob/master/js2-mode.el#L12387

PR #159, opened 2014-09-07
Add a jump to definition command bound to M-.

ScottyB:
I'm starting to split https://github.com/ScottyB/ac-js2 into separate packages and this navigation command is probably better suited to `js2-mode` than a completion package. I'm happy to sign any copyright forms for this pr. 

Thanks for your work on `js2-mode`!


dgutov:
@ScottyB

Here's one problem I've found: sometimes (or quite often, actually) when the code can't figure out the proper jump target, it seems like it chooses to jump to the beginning of the current node instead.

For instance, with this snippet:

```js
jQuery.fn.attr.call(selector, context, true);
```

Pressing `M-.` when on `attr` yields the expected "No jump location found". Whereas if point is on `call`, it will move to the beginning of the line instead. Could you take care of it?

ScottyB:
@dgutov,

That extra commit should take care of it. In regards to your other suggestion of letting the user choose between tags of the same name I am thinking of just cycling between candidates. What are your thoughts?

dgutov:
@ScottyB I'd rather not. This is the model used by `find-tag`, and a lot of users don't particularly like it.

The new navigation package `xref` (added to Emacs master not too long ago) uses a buffer to display all matches, and it's intended to replace `find-tag` and make `tags-loop-continue` obsolete.

ScottyB:
@dgutov That sounds much better! I guess it will be the same when searching in multiple buffers in which case this functionality is more like "find references" than "jump to definition".

dgutov:
I don't think multiple buffers change much.

Of course, when the logic is not precise the difference between them can get muddied, but "jump to definition" is ultimately about finding the one place (or the few) where this identifier was defined, and "find references" is about all of the places where it was referenced.

jacksonrayhamilton:
Variables, functions and parameters all work for me. I can't get object properties to work.

```bash
$ emacs --version
GNU Emacs 24.5.1
$ cd ~/projects/js2-mode/
$ emacs -Q -L . -l js2-mode.el
```

```lisp
var aObject = {prop1: 3, prop2: \"hello\"}; aObject.prop1
```

```
M-x js2-mode
```

I place point within the second instance of `aObject`, type `M-.` and get `Node is not a supported jump node`; same for the second instance of `prop1`. If change that to

```lisp
var aObject = {prop1: 3, prop2: \"hello\"};
aObject.prop1
```

then placing point over the second instance of `aObject` takes me to its definition, but placing point over the second instance of `prop1` yields `Symbol's function definition is void: first`.

ScottyB:
@jacksonrayhamilton These extra commits should address this issue let me know if you have any problems.

@dgutov Currently we jump to the last found location when there are multiple matches. Are you happy to merge this until Emacs 25 comes out and we can use the xref package?

jacksonrayhamilton:
Tried all the test cases manually, and they work.

js2-mode.el
```diff
@@ -12301,6 +12303,161 @@ it marks the next defun after the ones already marked."
...
+                     (right (js2-assign-node-right node))
+                     (temp-names names))
+                 (when (js2-prop-get-node-p left)
+                   (let* ((prop-list (js2-build-prop-name-list left))
+                          (prop-names (mapcar 'car prop-list))
+                          (found (loop for prop in prop-names
+                                       until (not (string= (pop temp-names) prop))
+                                       if (not temp-names) return prop)))
+
+                     ;; todo: clean this up!
```
jacksonrayhamilton: Can this todo be resolved?

tests/navigation.el
```diff
@@ -0,0 +1,60 @@
+;;; tests/navigation.el --- Some tests for js2-mode.
+
+;; Copyright (C) 2009, 2011-2013  Free Software Foundation, Inc.
```
jacksonrayhamilton: I believe this range should be "2009, 2011-2015"?

dgutov:
@jacksonrayhamilton Thanks for reviewing.

@ScottyB Thank you for the updates, the current version feels solid, and it's clearly a needed feature. Being able to use xref is orthogonal; it'll probably only be available for Emacs 25 anyway, and we have to keep supporting older versions.

The only minor concern I have left is a few of the new function names (`js2-names-left` sounds a bit... generic), but I'll deal with that myself.

dgutov:
@ScottyB Although, could you help with that as well?

- Can we replace `js2-build-prop-name-list` with using `js2-compute-nested-prop-get`?
- Please suggest a clearer name for `js2-names-left` and update its docstring for clarity (what's "list of names for a js2-prop-get-node"?)

ScottyB:
@dgutov Sure no problems. I hadn't noticed `js2-compute-nested-prop-get` so I was able to remove both `js2-build-prop-name-list` and `js2-names-left`. 

Did you want me to squash all the commits before you merge?

dgutov:
> Did you want me to squash all the commits before you merge?

No, it's okay.

dgutov:
Thanks again!

ScottyB:
That's ok, sorry it took so long to finish!

dgutov:
Sorry for the long review, too. ;-)

By the way, should we make this a minor mode? While the capability is neat, it doesn't work across files, so I imagine many users would still prefer using etags. And here we are shadowing its binding unconditionally.

ScottyB:
That is a fair point. I like the idea of navigation just working out of the box so perhaps a minor mode enabled by default is the way to go. This approach also forces etag users to "discover" the new functionality :).

dgutov:
Remapping `js-find-symbol` should be good enough. Although that command also has some advantages over `js2-jump-to-definition`: working across open buffers and jumping to any known symbol, with completion. Would be nice to see either, or both, implemented someday.

dgutov:
> working across open buffers

Oh, this is functional already; somehow it didn't work when I tried it previously. That's terrific.

ScottyB:
It should work, I tried it the other day and it was broken for object properties. I'll take a look and issue another pull request. Jumping to any known symbol current works using imenu, right?

dgutov:
> Jumping to any known symbol current works using imenu, right?

Right, but only within the current file.

> it was broken for object properties

Here's another example where `js2-jump-to-definition` is broken:

```js
var tee = {
  foo: function() {
    this.bar();
  },
  bar: function() {
  }
}
```

Trying to jump to `bar` in `this.bar` leads to a "Wrong type argument" error: the code can't cope with the beginning of the chain being not a `name-node` (but rather a keyword). It probably isn't too easy to fix: it would need to handle all the ways one can define a "class" in JavaScript. Or at least the most popular ones.

Issue #139, opened 2014-04-08
es6 support


zkat:
Harmony is adding a bunch of constructs, not just new keywords, and it would be nice to at least be able to use js2-mode. Specifically, some new keywords are highlighted as invalid and constructs like generator functions (`function*(){}`),  fat arrow functions (`a => a*a`), array and generator comprehensions, `export function() {...}`, etc, completely break the linter.

PR #175, opened 2014-11-04
ES6 module syntax support

cowboyd:
Provisional support for es6 module import/export syntax.

Because we use this syntax is a growing majority of our projects, I had to implement it. Obviously this doesn't track a spidermonkey implementation that doesn't yet exist, but for those that want provisional support, it could prove very useful (per the discussion in #139)

I'd love any feedback.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -1524,14 +1524,8 @@ the correct number of ARGS must be provided."
 (js2-msg "msg.mod.export.decl.at.top.level"
          "SyntaxError: export declarations may only appear at top level")
 
-(js2-msg "msg.mod.no.binding.name"
-         "SyntaxError: missing binding name")
-
 (js2-msg "msg.mod.rc.after.export.spec.list"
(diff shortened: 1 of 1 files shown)


theefer:
This is awesome, a massive thank you to both @cowboyd and @dgutov! I've been using ES6 with js2-module for a while now, and import/export support was the main frustration until now. Really happy to see this go in, many thanks!

cowboyd:
Yay! Well now that that's over and we're 100% guaranteed bug free for eternity, curious to know if you run into any problems.

theefer:
I was wondering, is it correct that `from` in the import syntax is not highlighted as a keyword?
```
import foo from 'bar';
           ^^^^
           not highlighted as a keyword like import, for, in, etc.
```

cowboyd:
It seems like it ought to be now that you mention it, but I didn't change anything related to keywords because I'm not sure what the full policy is on them.

PR #193, opened 2015-01-06
Add js2-normal-indent-hook.

XeCycle:
Used to alter indentation behavior without bounce indentation.

This allows the user to customize indentation without changing js2-mode or enabling bouncing mode (which is destructive).


dgutov:
Nope:

1. We're going to switch to reusing `js-mode` indentation code, so this request would belong there.

2. You can just advice `js2-proper-indentation`. No real need for the hook.

By the way, you've broken the build.

PR #195, opened 2015-01-09
js2-pretty-multiline-declarations: Add 'dynamic option

jacksonrayhamilton:
I liked how `js2-mode` indented multiline `var` statements, for the most part. But when it came to function expressions and object/array literals, I felt that the settings fell short.

With `(setq js2-pretty-multiline-declarations t)`, I got this:

```js
var a = function () {
    // Body.
};
```

But I also got this:

```js
var a = function () {
    // Body.
},
    b = function () {
        // Body.
    };
```

I did not like how the function bodies didn't align nicely.

So when I tried `(setq js2-pretty-multiline-declarations 'all)`, I was happy to find that I got this:

```js
var a = function () {
        // Body.
    },
    b = function () {
        // Body.
    };
```

But then I got this too:

```js
var a = function () {
        // Body.
    };
```

If there isn't another function in the `var` declaration to align against, the indentation seems superfluous, and I would rather not do it.

In the interest of maintaining backwards-compatibility for anyone who likes the `t` or `'all` behaviors, I have added a new property, `'dynamic`.

With `(setq js2-pretty-multiline-declarations 'dynamic)`, I get this:

```js
var a = function () {
    // Body.
};
```

And this:

```js
var a = function () {
        // Body.
    },
    b = function () {
        // Body.
    };
```

Which is exactly what I wanted. This PR provides that functionality.

I filled out and submitted a copyright assignment form two weeks ago, requesting to contribute another plugin of mine to `savannah.gnu.org/projects/emacs/`. I am not sure if I need to fill out a separate form to contribute to this repository, but I am willing to do so if needed.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -263,14 +263,37 @@ js2-mode also binds `js2-bounce-indent-backwards' to Shift-Tab."
       b = 20,
       c = 30;
 
-If the value is not `all', and the first assigned value in
+If the value is t, and the first assigned value in the
 declaration is a function/array/object literal spanning several
 lines, it won't be indented additionally:
 
   var o = {                   var bar = 2,
     foo: 3          vs.           o = {
   },                                foo: 3
-      bar = 2;                    };"
+      bar = 2;                    };
+
+If the value is `all', it will always be indented additionally:
+
+  var o = {
+        foo: 3
+      };
+
+  var o = {
+        foo: 3
+      },
+      bar = 2;
+
+If the value is `dynamic', it will be indented additionally only
+if the declaration contains more than one variable:
+
+  var o = {
+    foo: 3
+  };
+
+  var o = {
+        foo: 3
+      },
+      bar = 2;"
   :group 'js2-mode
   :type 'symbol)
 (js2-mark-safe-local 'js2-pretty-multiline-declarations 'symbolp)
@@ -10780,6 +10803,40 @@ In particular, return the buffer position of the first `for' kwd."
       (goto-char for-kwd)
       (current-column))))
 
+(defun js2-maybe-goto-declaration-keyword-end (bracket)
+  "Helper function for `js2-proper-indentation'.
+Depending on the value of `js2-pretty-multiline-declarations',
+move point to the end of a variable declaration keyword so that
+indentation is aligned to that column."
+  (cond
+   ((eq js2-pretty-multiline-declarations 'all)
+    (when (looking-at js2-declaration-keyword-re)
+      (goto-char (1+ (match-end 0)))))
+   ((eq js2-pretty-multiline-declarations 'dynamic)
+    (let (declaration-keyword-end
+          at-closing-bracket-p
+          comma-p)
+      (when (looking-at js2-declaration-keyword-re)
+        ;; Preserve the match data lest it somehow be overridden.
+        (setq declaration-keyword-end (match-end 0))
+        (save-excursion
+          (goto-char bracket)
+          (setq at-closing-bracket-p
+                ;; Handle scan errors gracefully.
+                (condition-case nil
+                    (progn
+                      ;; Use the regular `forward-sexp-function' because the
+                      ;; normal one for this mode uses the AST.
+                      (let (forward-sexp-function)
+                        (forward-sexp))
+                      t)
+                  (error nil)))
+          (when at-closing-bracket-p
+            (js2-forward-sws)
+            (setq comma-p (looking-at-p ","))))
+        (when comma-p
+          (goto-char (1+ declaration-keyword-end))))))))
+
 (defun js2-proper-indentation (parse-status)
   "Return the proper indentation for the current line."
   (save-excursion
@@ -10823,9 +10880,7 @@ In particular, return the buffer position of the first `for' kwd."
                                 (looking-at ")"))
             (backward-list))
           (back-to-indentation)
-          (and (eq js2-pretty-multiline-declarations 'all)
-               (looking-at js2-declaration-keyword-re)
-               (goto-char (1+ (match-end 0))))
+          (js2-maybe-goto-declaration-keyword-end bracket)
           (setq indent
                 (cond (same-indent-p
                        (current-column))
--- a/tests/indent.el
+++ b/tests/indent.el
@@ -112,3 +112,30 @@
   |  are kept
   |        unchanged!`"
   :keep-indent t)
+
+(js2-deftest-indent no-multiline-decl-first-arg-function-dynamic
+  "var foo = function() {
+  |  return 7;
+  |};"
+  :bind ((js2-pretty-multiline-declarations 'dynamic)))
+
+(js2-deftest-indent multiline-decl-first-arg-function-indent-dynamic
+  "var foo = function() {
+  |      return 7;
+  |    },
+  |    bar = 8;"
+  :bind ((js2-pretty-multiline-declarations 'dynamic)))
+
+(js2-deftest-indent multiline-decl-first-arg-function-indent-dynamic-comment
+  "var foo = function() {
+  |      return 7;
+  |    }/* MUAHAHAHA, ah ha! */,
+  |    bar = 8;"
+  :bind ((js2-pretty-multiline-declarations 'dynamic)))
+
+(js2-deftest-indent multiline-decl-first-arg-function-indent-dynamic-scan-error
+  "var foo = function() {
+  |  return 7;
+  |  ,
+  |  bar = 8;"
+  :bind ((js2-pretty-multiline-declarations 'dynamic)))


dgutov:
A few points:

1) Using the AST in the indentation code is bad (too slow). If you have a decently-sized file, a reindent-buffer command might not complete even when you come back from the coffee break. And you don't really have to, for this functionality. Just skip over the pair of curlies or brackets, and see if there's a comma after them (maybe separated with space/tab/newline/comment, though).

2) We very much want to offload the indentation logic to `js-mode` (the default JS mode in Emacs). It already includes `js--multi-line-declaration-indentation`, which only handles the `t` case (so there's no custom variable). It would be great if you could submit a patch to Emacs (against `master`) adding this new feature as an option.

3) In an existing file, this approach looks best to me too, but it has a problem (which was brought up in previous related discussions): if you're just writing the first function, line by line, there's no second function after it (there may even not be the closing curly). So what will you do when the second function is written? Go back and reindent? Some sane solution for this would be great, too (but not a requirement).

jacksonrayhamilton:
1) I did some dirty comparisons between the mode before and after my change. Indentation is already quite slow for large files. On my machine, it takes about 10 seconds to indent [jquery.js](http://code.jquery.com/jquery-2.1.0.js); after my addition, and with the new option enabled, it does not appear to be any slower. As for skipping, I need to ignore comments and strings and make sure to actually match the curlies if there are more functions or objects literals nested inside the function. I suppose it's possible, but that seems like a job that the parser has already handled, so implementing a mini-parser just for indentation doesn't seem worthwhile, or even much faster. Thoughts?

2) Depends on 1).

3) I think the `t` and `'all` options already cover this; if you stick to one of those styles, then you don't need to reindent. `'dyanmic` is for people with OCD who don't mind reindenting if it makes it look "right."

dgutov:
> after my addition, and with the new option enabled, it does not appear to be any slower.

Only if the file is already indented right. If not, it'll have to be reparsed after each modified line.

jacksonrayhamilton:
Okay. I'll see what I can do.

dgutov:
> As for skipping, I need to ignore comments and strings

Only whitespace and comments, you can use `js2-forward-sws` for this. And as for "actually matching the curlies", the indentation code is currently well-aware where the function's body starts (to skip to the closing, use `forward-sexp`, but handle scan errors).

> so implementing a mini-parser just for indentation doesn't seem worthwhile, or even much faster

It's orders of magnitude faster, and it's been implemented already.

jacksonrayhamilton:
Gotcha. Thanks for the tip, it should help.

jacksonrayhamilton:
How's 9b2cfd8 looking? By the looks of it, I think this could be ported to `js-mode` pretty easily.

dgutov:
Looks pretty good. Could you try to rewrite the patch against `js-mode` and submit it to Emacs, then?

jacksonrayhamilton:
Fixed.

Also, I said that I filled out the copyright assignment form, but I still haven't received word on its acceptance, and I'm still waiting for approval on the savannah `emacs` group. But I'd be happy to add this to `js.el` once I'm granted access.

dgutov:
Looks good.

> I still haven't received word on its acceptance

If you're not doing it electronically (by the virtue of living in US or Germany), the process can take several weeks, sometimes over a month. That's normal, unfortunately.

> But I'd be happy to add this to js.el once I'm granted access.

Why don't you send the patch with `M-x report-emacs-bug`? As a new committer, you could probably use a review anyway.

jacksonrayhamilton:
Would it be appropriate to submit a new feature as a "bug report?"

I don't think I will have to; just received word that my paperwork was accepted. I expect I'll be added to the memberlist soon.

Though if you'd like to take a peek, I've already ported it; seems to work:

```patch
commit cd62eb1d3ea3203b8bcc5dd28f209e87d66ed65e
Author: Jackson Ray Hamilton <jackson@jacksonrayhamilton.com>
Date:   Thu Jan 15 00:18:55 2015 -0800

    Add dynamic declaration indent option
    
    * js.el (js--proper-indentation): Add
    js-dynamic-multi-line-declaration-indent option.

diff --git a/lisp/progmodes/js.el b/lisp/progmodes/js.el
index c25e52c..17658f6 100644
--- a/lisp/progmodes/js.el
+++ b/lisp/progmodes/js.el
@@ -509,6 +509,37 @@ getting timeout messages."
   :type 'integer
   :group 'js)
 
+(defcustom js-dynamic-multi-line-declaration-indent nil
+  "Non-nil to indent the first function, array or object literal
+in a multi-line variable declaration specially.
+
+Normally, the first item is unindented, and subsequent items have
+their identifiers lined up against the first:
+
+  var o = {
+      foo: 3
+  };
+
+  var o = {
+      foo: 3
+  },
+      bar = 2;
+
+With this option enabled, the first declaration will be indented
+if there are more declarations, such that the first function,
+array or object lines up nicely with the remaining declarations:
+
+  var o = {
+      foo: 3
+  };
+
+  var o = {
+          foo: 3
+      },
+      bar = 2;"
+  :type 'boolean
+  :group 'js)
+
 ;;; KeyMap
 
 (defvar js-mode-map
@@ -1883,13 +1914,27 @@ In particular, return the buffer position of the first `for' kwd."
 	   ;; "case" and "default".
            (let ((same-indent-p (looking-at "[]})]"))
                  (switch-keyword-p (looking-at "default\\_>\\|case\\_>[^:]"))
-                 (continued-expr-p (js--continued-expression-p)))
+                 (continued-expr-p (js--continued-expression-p))
+                 declaration-keyword-end)
              (goto-char (nth 1 parse-status)) ; go to the opening char
              (if (looking-at "[({[]\\s-*\\(/[/*]\\|$\\)")
                  (progn ; nothing following the opening paren/bracket
                    (skip-syntax-backward " ")
                    (when (eq (char-before) ?\)) (backward-list))
                    (back-to-indentation)
+                   (when js-dynamic-multi-line-declaration-indent
+                     (and (when (looking-at js--declaration-keyword-re)
+                            ;; Preserve the match data lest looking-at overwrites it.
+                            (setq declaration-keyword-end (match-end 0)))
+                          (save-excursion (goto-char (nth 1 parse-status))
+                                          (when (condition-case nil
+                                                    (progn
+                                                      (forward-sexp)
+                                                      t)
+                                                  (error nil))
+                                            (while (forward-comment 1))
+                                            (looking-at ",")))
+                          (goto-char (1+ declaration-keyword-end))))
                    (let* ((in-switch-p (unless same-indent-p
                                          (looking-at "\\_<switch\\_>")))
                           (same-indent-p (or same-indent-p
```

dgutov:
> Would it be appropriate to submit a new feature as a "bug report?"

Entirely.

> I've already ported it; seems to work:

Looks okay to me. If you submit the patch, it might also get reviewed by someone else, though.

jacksonrayhamilton:
I've discovered an annoying edge case in this logic...

```js
var a = [
    'c',
    'd'
].join('\n'),
    b = {

    };
```

If the braces aren't immediately followed by a comma, then they won't be indented properly.

dgutov:
Indeed. I don't know if it's a big problem, but you can try handling the period case by doing some local parsing. Do a loop like this: 

See the period? Skip `js--dotted-name-re` forward. See open paren? Bind `forward-sexp-function` to nil and call `forward-sexp`. See the period again? Continue the loop. Otherwise, check for the comma.

Should look similar to `js--multi-line-declaration-indentation`, except shorter and easier to get right.

jacksonrayhamilton:
Bad news: `js2-mode-forward-sexp` calls `js2-reparse`, so the latest implementation would either be incorrect some of the time or would actually require a full parse (not just an interruptible "reparse") for each line.

I think parsing is going to quickly get out of hand here. There is more to this than just dot property access. Really we need to skip past an *expression* that spans multiple lines, and that could have virtually anything in it.

Is the strategy to choose an algorithm that's fast and gives accurate results for the majority of cases? Because in this scenario I'd rather have correct results *all* the time, even if it means some slowness in some rare situations. 

Another idea: It seems that the main issue with using the AST is that indentation mutates the buffer line-by-line, thus making the AST stale repeatedly. What do you think of the following algorithm?

- When indenting a line or region, when a `var` statement is encountered, and `js2-pretty-multiline-declarations` is `'dynamic`:
  - If the AST is stale (`js2-mode-buffer-dirty-p` is `t`?):
    - Call `js2-parse`.
  - Create an empty list in the variable `js2-virtual-indentation`.
  - After calculating the indentation for the current line, instead of actually indenting the line, `cons` the indentation onto `js2-virtual-indentation`.
  - For all remaining lines in the indented region:
    - When the previous line would be referred-to for indentation information, instead refer to the `car` of `js2-virtual-indentation`.
    - After calculating the indentation for the current line, instead of actually indenting the line, `cons` the indentation onto `js2-virtual-indentation`, as in the first case.
  - After reaching the end of the region, starting from the last line that should have been indented:
    - While js2-virtual-indentation` is non-nil:
      - Indent the current line by the `car` of `js2-virtual-indentation`.
      - Set `js2-virtual-indentation` to the `cdr` of `js2-virtual-indentation`.
      - If `js2-virtual-indentation` is non-nil, go to the previous line.

This has the advantage of keeping the AST accurate throughout the indentation process and reducing the maximum number of parses to 1. If the user wasn't power typing a moment ago then the number of parses will probably be 0. Supposing a modular codebase where files are usually no longer than 300 lines, you should never notice slowness, and supposing big files, you might only notice a slight hiccup sometimes when indenting the whole file or indenting some chunks (those which happen to have a declaration in them).

Any fatal flaws in that, or do you think there's a reasonable chance of you accepting a PR implementing this algorithm?

dgutov:
> Bad news: js2-mode-forward-sexp calls js2-reparse

It does. But do you actually need that? If not, just use the Lisp `forward-sexp` by binding `forward-sexp-function` to nil (like it's done in several places already).

> I think local parsing is going to quickly get out of hand here. There is more to this than just dot property access. Really we need to skip past an expression that spans multiple lines, and that could have virtually anything in it.

Right. You'd need something like `js--multi-line-declaration-indentation` in reverse. Note, though, that it's a relatively small function, and there are no currently known problems with it.

> Is the strategy to choose an algorithm that's fast and gives accurate results for the majority of cases?

That would be fine as well, I'd say.

> Another idea: It seems that the main issue with using the AST is that indentation mutates the buffer line-by-line, thus making the AST stale repeatedly. What do you think of the following algorithm?

Yes, it should work. At the cost of some extra complexity of sometimes referring to `js2-virtual-indentation` instead of `(current-column)`, and not doing that when we're just indenting one line.

But: `js-mode` doesn't have a JS parser, and I don't want to maintain a separate indentation logic without sufficient cause. I'm not convinced, so far.

jacksonrayhamilton:
I began implementing that algorithm, but came up with a much simpler approach.

I only planned on building up the `js2-virtual-indentation` list because I wanted to preserve the AST's freshness (by not mutating the buffer). But that seemed like it was going to be very inconvenient, as it looked like a lot of the indentation logic was intimately involved with the state of the current buffer.

So instead, I copied the contents of the buffer to a temporary buffer right before I made the AST stale. Then I indented (mutated) the buffer line-by-line, as normal. But when I needed to use the AST, I used the copy of the buffer to map my position (which might be at a different column number than before, due to indentation) back to the corresponding position in the stale AST. I still managed to keep the range of parses between 0 and 1, but with this approach I did not have to change any other indentation logic.

You can see the implementation [here](https://github.com/mooz/js2-mode/compare/2bb4db0898da30a149fa656dd4a68cb7e670e785...jacksonrayhamilton:d5893b680c5c248b37b1b575b7a815b5c32587a9).

It is rather painful to edit the first element of a `var` statement in a very large file, but for regular-sized files I don't notice a lag.

That said, I will admit the implementation in this PR's branch is probably sufficient for most cases.

dgutov:
That's pretty clever, but I believe you can do one better: instead of keeping a second buffer, just save the offset difference for each region line as you reindent it, and use those offsets whenever you need to call `js2-node-at-point`.

Either way, this scheme would be a big improvement for an indentation engine that uses the AST instead of `syntax-ppss`, like `js3-mode` has. If someone ported it to `js2-mode`, it could be a nice alternative.

dgutov:
@jacksonrayhamilton Okay, now that this feature is in `js-mode`, I guess we can merge it here to.

Could you resolve the current merge conflicts? Maybe rebase it on top of master, and squash the commits.

jacksonrayhamilton:
Resolved. Also, I refactored the code to look more like what is in `js.el`, so it should be easier to hand off to `js-mode` when the time comes.

dgutov:
Thanks!

PR #202, opened 2015-01-27
Add animationFrame functions

brianloveswords:
This adds two functions to the browser externs list:

- `requestAnimationFrame`
- `cancelAnimationFrame`

See http://www.w3.org/TR/animation-timing/#requestAnimationFrame for
details.


dgutov:
Thanks!

Issue #218, opened 2015-02-28
Error running timer `js2-mode-idle-reparse'


jacksonrayhamilton:
The latest js2-mode is throwing an error with the following buffer:

```
"use strict";
// comment
```

```
Error running timer `js2-mode-idle-reparse': (args-out-of-range 1 26)
```

Note that there is *not* a newline after the comment.

I tried to trace down the source of the error with `edebug-defun`, but farthest I got was what appeared to be the invocation of `run-hooks` at [line 7755](https://github.com/mooz/js2-mode/blob/90a905698cebcb9ce3adae86ab4d6733c511c831/js2-mode.el#L7755). At that point I used `C-h v js2-post-parse-callbacks` and found that it was `nil`, so I'm not sure what's up. Maybe it's related to the lack of the newline, because I don't get errors when there is a newline.

I get this error when running emacs with the flags `-batch -q --no-site-file`.

```
emacs --version
GNU Emacs 24.4.1
```

dgutov:
> because I don't get errors when there is a newline

I don't see the error either way, sorry. Are you sure no other modes were enabled when you tried?

> I get this error when running emacs with the flags -batch -q --no-site-file

Err, are you sure? The `-batch` argument would make it run in non-interactive mode (no interface), and then it would be pretty hard for you to try to reproduce this problem. Try `-Q` instead of all these arguments.

If you still see the error, `M-x toggle-debug-on-error`, and then evaluate `(js2-reparse t)`. This will give us a backtrace, at least.

jacksonrayhamilton:
Yes, I should have tried to reproduce it with `emacs -Q` first...

The error was coming at me twofold. It seemed be emerging from `context-coloring-mode`, which was one of those `js2-post-parse-callbacks`. So with the following buffer (no newline):

```js
//
```

I quickly pressed space and backspace, I got the following stack trace:

```lisp
Debugger entered--Lisp error: (args-out-of-range 1 4)
  add-text-properties(1 4 (face context-coloring-level-0-face))
  context-coloring-colorize-region(1 4 0)
  (let ((start (js2-node-abs-pos node))) (context-coloring-colorize-region start (+ start (progn (or (and (vectorp node) (>= (length node) 6) (memq (aref node 0) cl-struct-js2-node-tags)) (error "%s accessing a non-%s" (quote js2-node-len) (quote js2-node))) (aref node 3))) level))
[...]

PR #219, opened 2015-02-28
Add js2-declare-variable

jacksonrayhamilton:
I have a little present for you: A new utility function, `js2-declare-variable`!

`js2-declare-variable` makes it easy to extend `var` statements. Bound to C-c C-v, just enter the name of a variable (and optionally its initial value) and `js2-declare-variable` will append the declaration to the first `var` statement in the scope enclosing point.

If there isn't already a `var` statement in the enclosing scope, one will be created.

If there is a `"use strict"` statement or comments at the top of the enclosing scope, the `var` statement will be inserted after them.

In implementing this, I discovered a bug in `js2-node-get-enclosing-scope` where a node without a parent would trigger an error, because `js2-node-parent` could not operate on `nil`. The new fuction requires `js2-node-get-enclosing-scope` to behave correctly, so I fixed it.

I also noticed #218 when writing the `js2-declare-variable-comment-*` tests, so I simply added an extra newline in those tests to prevent that particular error. I would rather they not be there, so I marked them with TODOs until that bug is fixed. (**Update:** e054673 and efdc5f6 fixed that.)

I have signed the copyright forms for Emacs and they have been approved; I guess that means I can legally contribute to this repository as well?

I hope you like this feature and thanks for your review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,4 +23,5 @@ js2-imenu-extras.elc: js2-mode.elc
 
 test:
 	${EMACS} $(BATCHFLAGS) -l js2-mode.el -l tests/parser.el\
-	  -l tests/indent.el -l tests/externs.el -f ert-run-tests-batch-and-exit
+	  -l tests/indent.el -l tests/externs.el -l tests/utilities.el\
+	  -f ert-run-tests-batch-and-exit
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -6169,8 +6169,6 @@ its relevant fields and puts it into `js2-ti-tokens'."
                              (setf (js2-token-beg token) (- js2-ts-cursor 2))
                              (js2-skip-line)
                              (setf (js2-token-comment-type token) 'line)
-                             ;; include newline so highlighting goes to end of window
-                             (cl-incf (js2-token-end token))
                              (throw 'return js2-COMMENT))
                            ;; is it a /* comment?
                            (when (js2-match-char ?*)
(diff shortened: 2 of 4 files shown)


jacksonrayhamilton:
This PR now depends on [\#220](https://github.com/mooz/js2-mode/pull/220).

dgutov:
I can see quite a bit of effort went into this, but are you sure `js2-mode` is the best place for it, as opposed to [js2-refactor](https://github.com/magnars/js2-refactor.el)? It implements a set of editing commands already, and this one might be a welcome addition.

> I have signed the copyright forms for Emacs and they have been approved; I guess that means I can legally contribute to this repository as well?

It does. Magnar also signed it, so it probably means we could add js2-refactor to GNU ELPA without too much effort, in case that's your primary interest in contributing this command here.

jacksonrayhamilton:
Seeing as [\#159](https://github.com/mooz/js2-mode/pull/159) is pending I assumed this would be an okay place to put it.

I'd have to rewrite the tests, but if you think it belongs there, I can suggest it to the maintainer. We don't necessarily need to get the package on ELPA, do we? I can just as easily use MELPA.

dgutov:
> Seeing as #159 is pending I assumed this would be an okay place to put it.

That pull request is different: it's navigation, not editing. I don't think we actually have any editing commands here. And we add one, the other ones must follow, and then we'll start duplicating `js2-refactor`.

> I'd have to rewrite the tests, but if you think it belongs there, I can suggest it to the maintainer.

Let's start with asking him. Maybe you won't have to rewrite the tests that much.

> We don't necessarily need to get the package on ELPA, do we?

Sure, it's up to you. ELPA just means a wider audience.

jacksonrayhamilton:
I've continued development of this feature in `js2-refactor` and the author seems interested. Thanks for pointing me in the right direction.

PR #220, opened 2015-02-28
Fix off-by-one error for comment node length

jacksonrayhamilton:
Fixes [\#218](https://github.com/mooz/js2-mode/issues/218).

See [here](https://github.com/mooz/js2-mode/issues/218#issuecomment-76540657) for more info.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -6163,8 +6163,11 @@ its relevant fields and puts it into `js2-ti-tokens'."
                              (setf (js2-token-beg token) (- js2-ts-cursor 2))
                              (js2-skip-line)
                              (setf (js2-token-comment-type token) 'line)
-                             ;; include newline so highlighting goes to end of window
-                             (cl-incf (js2-token-end token))
+                             ;; include newline so highlighting goes to end of
+                             ;; window, if there actually is a newline; if we
+                             ;; hit eof, then implicitly there isn't
+                             (unless js2-ts-hit-eof
+                               (cl-incf (js2-token-end token)))
                              (throw 'return js2-COMMENT))
                            ;; is it a /* comment?
                            (when (js2-match-char ?*)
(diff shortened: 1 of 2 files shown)


js2-mode.el
```diff
@@ -6163,8 +6163,6 @@ its relevant fields and puts it into `js2-ti-tokens'."
                              (setf (js2-token-beg token) (- js2-ts-cursor 2))
                              (js2-skip-line)
                              (setf (js2-token-comment-type token) 'line)
-                             ;; include newline so highlighting goes to end of window
```
dgutov: Including the newline is somewhat important actually (syntax-based font-locking does that).

I think this piece of code should look if there's a newline after, and if so, include it.

jacksonrayhamilton: But then it's not the correct length, it's just a convenient length.

Could the +1 calculation be done at the time of font-locking when a node type is found to be a comment?

dgutov: It's semantically meaningful too: `//` comments stretch until the end of the line, so it makes certain amount of sense to include the newline.

And doing it at highlighting time is inconvenient: see the end of `js2-get-token-internal`, special-casing `js2-COMMENT` there is inconvenient, moreso because line and block comments get the same token type.

jacksonrayhamilton: > `//` comments stretch until the end of the line

That does make sense.

Okay, I will try special-casing for comment nodes elsewhere.

dgutov: Err, that's not what I suggested. The piece of code we're discussing right now should be fixed, to usually, but not always, include the newline.

jacksonrayhamilton: Yes, just re-read your first comment, and I agree with that too. I will add some cases for that.

jacksonrayhamilton:
How's this look?

dgutov:
Looks good, thanks. Could you squash and force-push?

jacksonrayhamilton:
Done.

PR #221, opened 2015-03-01
Fix js2-node-get-enclosing-scope

jacksonrayhamilton:
Previously this function discriminated against nodes that did not have parents. For such nodes, `nil` should have been returned, but instead an error was emitted, because a parentless node would produce the initial parent `nil` which would cause `js2-node-parent` to blow up. This fixes that.

Originally this was going to be a throw-in for [\#219](https://github.com/mooz/js2-mode/pull/219), but considering we might not merge that in, I'd like to get this fix in so I can move forward on a `js2-refactor` version of [\#219](https://github.com/mooz/js2-mode/pull/219).


dgutov:
Thanks!

Issue #230, opened 2015-04-22
ES6 computed property names for methods


ef4:
I found some valid ES6 that doesn't parse:

```js
class MyClass {
  [ 'say' + 'Hello']() {
    return "hello";
  }
}

var x = {
  [ 'say' + 'Hello']() {
    return "hello";
  }
 }
```

Using 20150406.1341 from Melpa.

dgutov:
@shicks Could you take care of this?

shicks:
Okay, I'll take a look.

shicks:
See pull request #231

PR #231, opened 2015-04-27
Generalize object literal parsing to understand computed function names.

shicks:
This allows e.g. {[a + b]() {...}}, {'foo'() {...}}, and {get [a + b]() {...}}.

Previously, short-hand functions/getters/setters/generators only worked for unquoted names.  Now we support all combinations of key types (name, string, number, computed) with all property types (colon-delimited, getter, setter, generator, and short-hand functions).

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -3952,6 +3952,7 @@ and both fields have the same value.")
 
 (defun js2-print-object-prop-node (n i)
   (let* ((left (js2-object-prop-node-left n))
+         (right (js2-object-prop-node-right n))
          (computed (not (or (js2-string-node-p left)
                             (js2-number-node-p left)
                             (js2-name-node-p left)))))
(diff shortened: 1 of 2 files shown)


shicks:
This is in response to Issue #230

ef4:
I can confirm this solves me issue. Thanks!

js2-mode.el
```diff
@@ -10499,7 +10525,7 @@ and expression closure style is also supported
 
 POS is the start position of the `get' or `set' keyword.
 PROP is the `js2-name-node' representing the property name.
-GET-P is non-nil if the keyword was `get'."
+TYPE-STRING is a string `get', `set', or `', indicating a found keyword."
```
dgutov: Was the last option meant to be an asterisk? I think nil is also a possible value.

shicks: Yes, fixed.  I also fixed a minor syntax highlighting issue in js2-parse-named-prop where it was highlighting the [ and ] differently for computed shorthand functions, which was weird.

dgutov:
Thanks!

PR #232, opened 2015-05-03
Don't unset `js2-recorded-identifiers'

jacksonrayhamilton:
Forgive me if I'm missing something, but I don't see why this variable is being unset.

It would be useful if it was still set, especially if a minor mode wanted to analyze the identifiers from `js2-mode`'s initial parse. Before this change, a minor mode would have to set up a `js2-post-parse-callback` and reparse for no reason in order to get access to the `js2-recorded-identifiers`.


dgutov:
Indeed, there seems to be no good reason for that. We set it to nil in `js2-do-parse` anyway.

Thanks.

Issue #212, opened 2015-02-07
Unused variables


lelit:
How difficult would be to implement unused variables recognition?

dgutov:
Shouldn't be too hard. It could reuse the mechanism that tracks undeclared identifiers.

lelit:
I've tried to implement the feature, and to my surprise it works, more or less :yum: 
It's far from being good enough, but maybe you can take a look and suggest a better way, or style, or whatever? Thank you in advance for *any* hint!

https://github.com/lelit/js2-mode/commit/8fcee0f52ff93ce2ad09bfebb7a8a4a2c4d3486b

lelit:
I've refined one of the functions, and tested it against common cases.
See https://github.com/lelit/js2-mode/commit/9ff8ee5f23d3387c1e027f41b91abd6cd2fa1b59

dgutov:
Thanks for this, but you're mixing two approaches.

Using `js2-visit-ast`, you could make this a third-party package (or at least a separate mode), that would add an element to `js2-post-parse-callbacks`, and visit `js2-mode-ast`.

But if you're making this a part of the parser itself, you can try to directly hook into some relevant functions. Looks at the two places that call `js2-record-name-node`: these parse variable references. Maybe we could consider them all "usages". And for declarations themselves, maybe hooks into `js2-define-new-symbol`.

I have no strong preference between these approaches, but it would be good to try the first one, to see if it's fast enough.

lelit:
Yes, thank you, I'm learning as I go, I will try the *extension* method...

Do you have an explanation why the current `js2-highlight-undeclared-vars` is implemented that way, that is one first pass that uses `js-record-name-node` to keep track of used names, and then one final loop that actually decorates them, instead of a simple traverse using `js2-visit-ast` taking advantage of the `end-p` flag to discriminate the point where all the children have been visited (similar to the approach I taken in my [js2-highlight-unused-variables](https://github.com/lelit/js2-mode/blob/issue212/js2-mode.el#L7091))?

dgutov:
> Do you have an explanation why the current js2-highlight-undeclared-vars is implemented that way

Two reasons: It was halfway implemented that way before me already. And using `js2-visit-ast` is almost certain to be slower (but by how much?).

But I've always entertained the idea of trying to do that the other way, because it's conceptually cleaner.

lelit:
[...]

PR #233, opened 2015-05-07
Issue212

lelit:
This branch implements unused variable detection (see #212), enabled by a minor mode, `js2-highlight-problematic-variables-mode`.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -10372,7 +10372,7 @@ We should have just parsed the 'for' keyword before calling this function."
           (while (js2-match-token js2-IF)
             (push (car (js2-parse-condition)) filters))
           (setq expr (js2-parse-assign-expr))
-          (setq first (car (last loops))))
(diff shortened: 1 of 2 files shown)


dgutov:
Here's a couple of issues before I merge this:

* We always classify the error variable (in a catch clause) as "referenced but never initialized".

* Could we find a better name than `js2-highlight-problematic-variables-mode`? That's pretty vague. Undeclared variables are also problematic, yet it doesn't touch those. I think `js2-highlight-unused-variables-mode` would be better, even if it also warns about uninitialized variables.

lelit:
Sure, will do!

lelit:
Let me know if last two commits satisfy your request, thank you!

dgutov:
Looks good. Thank you for the contribution. :thumbsup:

lelit:
I learned quite a lot on another side of Emacs :) Thank you!

Issue #185, opened 2014-12-04
Async / Await


arcanis:
Async and await are considered for inclusion in ES7. Even if they are not yet in Spidermonky, since it seems to be mainly a matter of adding two keywords, could it be possible to support them ? Traceur is already able to transpile it.

```js
var foo = async function ( ) {
    await bar( );
};

async function foo( ) {
    await bar( );
};

class Foo {
    async function foo( ) {
        await bar( );
    }
}
```

mooz:
Implemented `async`/`await` based on the proposal, and pushed to https://github.com/mooz/js2-mode/tree/es7-async-await branch. Since [`6to5`](https://github.com/6to5/6to5), a quite popular ES6/7 to ES5 transpiler, already supports `async`/`await` (experimentally), I think it's good if `js2-mode` also supports it.

If there is no problem, I'll merge the branch to `master`.

dgutov:
@arcanis Do you have a link that actually describes the changes to the grammar? Are `async` and `await` supposed to be new "proper" keywords, or contextual keywords? @mooz changed the implementation for the latter, but if they turns out to be the former, it will be unnecessary complexity.

mooz:
@dgutov @arcanis

According to ES6 draft and tc39's discussion for ES7,

- `async` won't be reserved (is a contextual keyword) (https://github.com/rwaldron/tc39-notes/blob/master/es6/2014-11/nov-18.md#413-add-async-as-futurereservedword) [tc39's discussion],
- `await` is a future reserved word only if a program is parsed as a module (?) (https://people.mozilla.org/~jorendorff/es6-draft.html#sec-future-reserved-words)[spec].

arcanis:
@mooz Your commits work fine, except for class methods (which should work according to [Traceur](http://google.github.io/traceur-compiler/demo/repl.html#%2F%2F%20Options%3A%20--async-functions%20%0A%0Aclass%20Foo%20%7B%0A%20%20%0A%20%20static%20async%20myFunc()%20%7B%0A%20%20%7D%0A%0A%20%20async%20myFunc()%20%7B%0A%20%20%7D%0A%0A%7D%0A))

amasad:
n00b question. How do I install this from the branch?

arcanis:
@mooz Do you think you could support the `class { async foo() { } }` syntax on your branch? I've looked, but I'm far from being fluent enough in lisp to understand what to change to be clean :(

mooz:
[...]

PR #265, opened 2015-09-11
Adapt async/await to latest master

XeCycle:
async/await support by @mooz has been there for 8 months.  There are some TODO around the lines, which I suppose he intended to improve later, but this has been too long.  Anyway it basically works, let's merge first and resolve problems later (if any).

Btw I guess my changes do not qualify as legally significant.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -610,7 +610,9 @@ which doesn't seem particularly useful, but Rhino permits it."
 (defvar js2-NO_SUBS_TEMPLATE 167) ; template literal without substitutions
 (defvar js2-TAGGED_TEMPLATE 168)  ; tagged template literal
 
-(defconst js2-num-tokens (1+ js2-TAGGED_TEMPLATE))
+(defvar js2-AWAIT 169)  ; await (pseudo keyword)
+
+(defconst js2-num-tokens (1+ js2-AWAIT))
 
 (defconst js2-debug-print-trees nil)
 
@@ -1354,6 +1356,9 @@ the correct number of ARGS must be provided."
 
 (js2-msg "msg.yield.parenthesized"
          "yield expression must be parenthesized.")
+
+(js2-msg "msg.bad.await"
+         "await must be in async functions.")
 
 ;; NativeGlobal
 (js2-msg "msg.cant.call.indirect"
@@ -3246,6 +3251,7 @@ a `js2-label-node' or the innermost enclosing loop.")
                                                           params rest-p
                                                           body
                                                           generator-type
+                                                          async
                                                           lp rp)))
   "AST node for a function declaration.
 The `params' field is a Lisp list of nodes.  Each node is either a simple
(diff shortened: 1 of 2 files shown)


dgutov:
Thank you.

> I guess my changes do not qualify as legally significant.

It appears so.

However, I'd really rather have a contributor who would make this feature their responsibility, especially since some details of the syntax could change as it moves through the standardization process. @XeCycle Are you interested in being that person?

Could you answer the comment about `static` + `async` in #185?

XeCycle:
@dgutov I answered that question.

I think I should be able to answer questions on the standard text, or forward the question to tc39 bug tracker.  But currently I am not yet familiar with the parser here in js2-mode, as it is constructed quite different from the grammar specified in the standards.  Did we intend it to be more permissive or for performance?

Well, if you assign this feature to me, expect roughly 1~2 weeks for my answer on questions.

dgutov:
>  But currently I am not yet familiar with the parser here in js2-mode, as it is constructed quite different from the grammar specified in the standards.

I wouldn't say it's too special: it's a recursive-descent parser, adapted from SpiderMonkey.

> Did we intend it to be more permissive or for performance?

Isn't it more or less the same? I'd say we want to only allow valid grammar, as long as it's not too much of a pain to implement, and doesn't slow us down too much.

> if you assign this feature to me, expect roughly 1~2 weeks for my answer on questions

I'd be happy to leave this fix to you, as long as you keep in mind the copyright assignment requirements.

XeCycle:
> I wouldn't say it's too special: it's a recursive-descent parser, adapted from SpiderMonkey.

I have not read ES<=3, no idea if they were different from the beginning or diverged later.  Some grammar productions are given different semantics and/or are defined differently, e.g. the left hand side of an assignment expression is defined by the standard to be a `LeftHandSideExpression` which expands to `NewExpression` or `CallExpression`, but we parse for `ConditionalExpression` and are only checking strict identifier if it is a `js2-name-node`.  This is too permissive.

There are other pieces where our code is permissive to bad code; the above example seems to be avoiding parse the left hand side a second time, so I cannot guess our current policy on standards conformance.

> I'd be happy to leave this fix to you, as long as you keep in mind the copyright assignment requirements.

What about just send me the forms?  I plan to add bind operator sooner or later, so I need it after all.

dgutov:
> I have not read ES<=3, no idea if they were different from the beginning or diverged later.

You could get a better idea by looking at the current SpiderMonkey sources, and comparing `js2-mode` with them. Can't say about this particular problem, but we're lagging behind SM in certain places. Although it could be buggy in this respect as well.

> so I cannot guess our current policy on standards conformance

Use your best judgment, I guess. You can assume that this codebase is not particularly well-maintained.

> What about just send me the forms?

Sure. http://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.future

Put in "Emacs" as "the name of the program or package you're contributing to".

XeCycle:
Now working on adding async method support to solve that comment about `static` + `async`.  For the names to be less confusing, I would like to change `js2-getter-setter-node` to `js2-method-node` and have it cover all five types of methods at http://tc39.github.io/ecmascript-asyncawait/#MethodDefinition.  So:

1. how do you like this change?
2. shall that be a separate PR?

btw request form sent, waiting for reply.

dgutov:
> I would like to change js2-getter-setter-node to js2-method-node

Sounds good. Please make it a separate PR, I'll merge it first.

XeCycle:
Taking on async methods, closing this.  Pls wait for another PR :)

PR #266, opened 2015-09-11
Don't treat getter / setter pairs as duplicates

### LINKED ISSUE #264: "Property 'x' already defined" for getters and setters and 'use strict'
It seems that getters and setters with the same name are being confused for the same property. The error message is only shown in strict mode.

For example, given the following code in a `js2-mode` buffer:

```javascript
'use strict';

class X {
  get x() {  }
  set x(val) {  }
}
```

The `x` in `set x(val) { }` will be highlighted red, and it will show the message:

`Property 'x' already defined in this object literal.`

jacksonrayhamilton:
Fixes #264.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -10654,7 +10654,13 @@ expression)."
                     (lambda (previous-elem)
                       (and (setq previous-elem-key-string
                                  (js2-property-key-string previous-elem))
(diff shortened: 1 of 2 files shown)


dgutov:
Thanks!

PR #271, opened 2015-10-06
Add JSX indentation support

### LINKED ISSUE #140: Highlighting & indenting embedded XML (E4X)?
Hello,

Please let me know.
Thanks!

jacksonrayhamilton:
Partially resolves #140.

Sorry, it's rather slow for large files, so I made it an opt-in feature. I am rather happy with the test cases. Let me know what you think!

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -11281,8 +11281,9 @@ Selecting an error will jump it to the corresponding source-buffer error.
   "\\(?:[[:space:]\n]\\|//.*\\|/\\*\\(?:.\\|\n\\)*?\\*/\\)*?"
   "Match whitespace and comments.")
 
-(defconst js2-jsx-start-tag-re
-  (concat "\\([(=:,]\\)" js2-whitespace-and-comments-re "\\(<\\)" sgml-name-re)
+(defconst js2-jsx-before-tag-re "[(=:,]")
+
+(defconst js2-jsx-start-tag-re (concat "<" sgml-name-re)
   "Find where JSX starts.
 Assume JSX appears in the following instances:
 - Inside parentheses, when returned or as the first argument to a function
@@ -11327,6 +11328,7 @@ Currently, JSX indentation supports the following styles:
   );"
   (let ((current-pos (point))
         (current-line (line-number-at-pos))
+        last-pos
         before-tag-pos before-tag-line
         tag-start-pos tag-start-line
         tag-end-pos tag-end-line
@@ -11337,11 +11339,17 @@ Currently, JSX indentation supports the following styles:
        ;; Determine if we're inside a jsx element
        (progn
          (end-of-line)
-         (re-search-backward js2-jsx-start-tag-re nil t))
+         (while (and (not tag-start-pos)
+                     (setq last-pos (re-search-backward js2-jsx-before-tag-re nil t)))
+           (goto-char (match-end 0))
+           (js2-forward-sws)
+           (when (looking-at js2-jsx-start-tag-re)
+             (setq before-tag-pos last-pos
+                   tag-start-pos (match-beginning 0)))
+           (goto-char last-pos))
+         tag-start-pos)
        (progn
-         (setq before-tag-pos (match-beginning 1)
-               before-tag-line (line-number-at-pos before-tag-pos)
-               tag-start-pos (match-beginning 2)
+         (setq before-tag-line (line-number-at-pos before-tag-pos)
                tag-start-line (line-number-at-pos tag-start-pos))
          (and
           ;; A "before" line which also starts an element begins with js, so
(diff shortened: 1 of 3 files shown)


js2-mode.el
```diff
@@ -11272,6 +11283,204 @@ Selecting an error will jump it to the corresponding source-buffer error.
...
+        ;; like sgml relative to the first thing
+        ((= current-line tag-start-line) 'first)
+        ('nth))))))
+
+(defmacro js2-as-sgml (&rest body)
+  "Execute BODY as if in sgml-mode."
+  `(with-syntax-table sgml-mode-syntax-table
+     (let (forward-sexp-function
+           parse-sexp-lookup-properties
+           (sgml-basic-offset js2-basic-offset))
```
dgutov: Are you sure the users will want JSX to be indented with the same offset as JS?

dgutov:
I'm happy it took a fairly moderate amount of code. But don't you think JSX indentation would be useful to support in `js-mode` as well?

Note that all JSX indentation tests fail in the current Emacs master (Travis just doesn't see it, because the only `emacs-snapshot` it can use is ancient). Which likely means they fail whenever `js2-old-indent` is not used.

js2-mode.el
jacksonrayhamilton: Pretty sure. JSX is described as an extension of JS, rather than as "embedded XML in JS". In most code snippets I've seen, the indentation people use is the same. Due to many editors binding TAB to "insert N whitespace characters," it's likely that other people on a team will use consistent indentation throughout JSX files.

Although, `sgml-attribute-offset` is not also rebound, and that could be anything.

So I see a few options here:
- Make a `js2-jsx-basic-offset` and/or `js2-jsx-attribute-offset`
  - Pros: Neither standard JS nor SGML preferences necessarily dictate this indentation
  - Cons: More variables, and the user likely(?) wants the offset to be the same
- Use `sgml-basic-offset` and `sgml-attribute-offset`
  - Pros: Fewer variables
  - Cons: Presumes that SGML preferences carry over to JS/JSX

dgutov: I think either is fine. Note that in the second case the user could still change the JSX indentation by changing `sgml-basic/attribute-offset` locally in `js2-mode-hook`.

jacksonrayhamilton:
> But don't you think JSX indentation would be useful to support in `js-mode` as well?

Sure, and it would be pretty easy to move this there.

Do you have any ideas on how to improve performance? Or is this good enough? (Line-by-line it seems instant, but indenting a 4000-line file takes about 1 minute on my machine.)

dgutov:
How long does indenting a 4000-line HTML file take?

dgutov:
Or do you mean indenting a 4000-line JS file (without JSX) also takes close to that time, as long as JSX indentation support is enabled?

jacksonrayhamilton:
> How long does indenting a 4000-line HTML file take?

Less than a second.

> Or do you mean indenting a 4000-line JS file (without JSX) also takes close to that time, as long as JSX indentation support is enabled?

Yes, I did mean the latter statement. The test file was [jslint.js](https://raw.githubusercontent.com/douglascrockford/JSLint/8d015dbd52e7f359bc25e0e5c05e129ef57c771e/jslint.js).

My guess is that the "`re-search-backward` and parse" algorithm is really inefficient because the entire file gets searched for each line, and there are lots of false positives (1896 occurences of `(`, 1979 of `=`, 455 of `:` and 1445 of `,`). Originally I used a more complex regexp, but that resulted in stack overflow at about 40% indentation. Maybe a happy medium would help (in the regexp, check for a space or `<` immediately after one of the "JSX incoming" characters, for instance), but that doesn't improve the underlying algorithm.

dgutov:
Before doing `re-search-forward` for the opening tag, can't you look for `js2-jsx-start-tag-re` on the current line first? Or on the previous line, if the current one is empty?

XeCycle:
https://github.com/jacksonrayhamilton/js2-mode/blob/jsx/js2-mode.el#L9833
https://github.com/jacksonrayhamilton/js2-mode/blob/jsx/js2-mode.el#L9885
I guess these lines tell where XML starts and ends.  What about placing a mark at these points?

dgutov:
@XeCycle We don't rely on the parser for indentation because that's usually much slower.

jacksonrayhamilton:
I ported this to `js-mode` and submitted a bug for review. In the process, I replaced the `re-search-backward` call with this:

```elisp
(defsubst js--jsx-find-before-tag ()
  (let (pos)
    (while (and (> (point) (point-min))
                (not (progn
                       (end-of-line 0)
                       (when (or (eq (char-before) 40)   ; (
                                 (eq (char-before) 44))  ; ,
                         (setq pos (1- (point))))))))
    pos))
```

Which resulted in a 62% performance gain for indenting [jslint.js](https://raw.githubusercontent.com/douglascrockford/JSLint/8d015dbd52e7f359bc25e0e5c05e129ef57c771e/jslint.js) with `js.el` byte-compiled (23.19s -> 8.91s on my machine).

I'm also now using `sgml-basic-offset` and `sgml-attribute-offset` for the JSXElements' offsets.

PR #273, opened 2015-10-13
Implement bind operators

XeCycle:
Bind operators as specified by https://github.com/zenparsing/es-function-bind.

This turns off E4X accessors separately because there is a conflict; simple XML literals as used by JSX is preserved.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -738,6 +738,7 @@ parser as a frontend to an interpreter or byte compiler.")
 (js2-deflocal js2-compiler-use-dynamic-scope nil)
 (js2-deflocal js2-compiler-reserved-keywords-as-identifier nil)
 (js2-deflocal js2-compiler-xml-available t)
+(js2-deflocal js2-compiler-e4x-accessor-available nil)
 (js2-deflocal js2-compiler-optimization-level 0)
 (js2-deflocal js2-compiler-generating-source t)
 (js2-deflocal js2-compiler-strict-mode nil)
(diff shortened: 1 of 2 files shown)


XeCycle:
Wait some time, some changes are needed:

- AST for `a::b()` is like `(a)::(b())`, but it should be like `((a)::(b))()`
- for the above reason, a statement like `a::b()` is considered to have no side effects

js2-mode.el
```diff
@@ -9913,7 +9915,44 @@ Returns the list in reverse order.  Consumes the right-paren token."
...
 
-(defun js2-parse-member-expr (&optional allow-call-syntax)
+(defun js2-parse-lhs-expr ()
+  "Parse a `LeftHandSideExpression'.
+
+The three possible productions are collectively handled like a
+`CallExpression' with optional `Arguments' at end.  Unary bind is
+handled specially because it is the only one not beginning with a
+`NewExpression'."
+  (let ((pn (if (= (js2-current-token-type) js2-COLONCOLON)
```
XeCycle: While this works, I wonder why it is `js2-current-token-type` not `js2-peek-token`.  What is the difference?

dgutov: Because the latter function looks at the *next* token?

XeCycle: I see; so this function, similar to the previous `js2-parse-member-expr`, expects _current token_ to be the start of its expression.  It is now called only from `js2-parse-unary-expr` which does that.  Seems different parse functions expect different preconditions, e.g. `js2-parse-assign-expr` begins with `(let ((tt (js2-get-token)) ...`.

dgutov: That's correct. Not sure if it's easy or beneficial to make this aspect more consistent.

dgutov:
What's the status of this proposal? Can we be confident that the syntax won't change?

XeCycle:
It is stage 0, and I am not that confident.  But this is one of the 3 implemented stage 0 proposals in babel https://babeljs.io/docs/usage/experimental/ (class properties entered stage 1), and there is already a library supporting its usage https://github.com/jussi-kalliokoski/trine.  I feel positive on this, but yes, further discussion needed.

dgutov:
Conflict with E4X is not that big a problem (I'm not aware of anyone using it either, except for JSX).

But since the proposal is not going ahead, this is a no-go.

PR #276, opened 2015-10-15
Rename getter-setter-node to method-node

XeCycle:
This is the PR for refactor as you wanted @dgutov.

Something to note:

- method-node is still an infix node, but its right, the function, is now named (more closely follows standard);
- its left, the property, is not a child of the method node, but instead a child of the right side function node (duplicated in the tree).

These allows `js2-infix-node-left` and `js2-infix-node-right` to be still able to access the key and value of a method node.  Another option is to make method-node inherit directly from the base node, and add new functions `js2-property-key` and `js2-property-value`.  Quite some functions need change; I am not sure if I can catch all of them by looking at test results.

Please tell whether you like the current implementation or the other option.  It may need careful reviews if going the latter.

If going the former, tell me if you'd like the last two commits rebased into one.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 * Object properties are highlighted using a different face:
   `js2-object-property`, which has no color by default.
+* `js2-getter-setter-node` is renamed to `js2-method-node`, together
+  with its related functions.  It already handles generator methods,
+  and will in the future add support for async methods, so the old
+  name will get more confusing.
 
 ## 20150909
 
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -3884,6 +3884,30 @@ optional `js2-expr-node'"
   (js2-print-list (js2-object-node-elems n))
   (insert "}"))
 
+(cl-defstruct (js2-computed-prop-name-node
+               (:include js2-node)
+               (:constructor nil)
+               (:constructor make-js2-computed-prop-name-node
+                             (&key
+                              (type js2-LB)
+                              expr
+                              (pos (js2-current-token-beg))
+                              (len (- js2-ts-cursor
+                                      (js2-current-token-beg))))))
+  "AST node for a `ComputedPropertyName'."
+  expr)
+
+(put 'cl-struct-js2-computed-prop-name-node 'js2-visitor 'js2-visit-computed-prop-name-node)
+(put 'cl-struct-js2-computed-prop-name-node 'js2-printer 'js2-print-computed-prop-name-node)
+
+(defun js2-visit-computed-prop-name-node (n v)
+  (js2-visit-ast (js2-computed-prop-name-node-expr n) v))
+
+(defun js2-print-computed-prop-name-node (n i)
+  (insert (js2-make-pad i) "[")
+  (js2-print-ast (js2-computed-prop-name-node-expr n) 0)
+  (insert "]"))
+
 (cl-defstruct (js2-object-prop-node
                (:include js2-infix-node)
                (:constructor nil)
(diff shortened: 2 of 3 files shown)


XeCycle:
Another problem came up in my mind: ES6 allows a `ClassElement` to be `;` which is not yet supported.

This PR did not include fix for that --- I assume you want such a refactor not to change its behaviour.  But when fixing that later, it will make even less sense to parse a `;` into an infix node.  Seems I have to either change method-node to inherit from base node, or split up object literal handling and class body handling, or both.  ES7 property initializers are less like infix nodes, so I suggest we do both, and extract dup check code out to serve both cases.

dgutov:
> its left, the property, is not a child of the method node, but instead a child of the right side function node (duplicated in the tree)

I don't really like that kind of duplication. Can you name a practical benefit for it?

> method-node is still an infix node, but its right, the function, is now named (more closely follows standard)

The standard described the syntax of the language, but it doesn't exactly prescribe the shape of AST.

> This PR did not include fix for that --- I assume you want such a refactor not to change its behaviour.

This is not really a refactoring (it changes the behavior), but granular pull requests are always better.

> But when fixing that later, it will make even less sense to parse a ; into an infix node.

Why? If the language simply allows optional semicolons between methods, we could simply skip over them, and not include them in the AST. It's not like they bring any semantic value, right?

XeCycle:
I checked method definitions again, and, sorry, my mistake.

A method definition creates a named function, but without the name referring to itself in its own scope; I had thought it to be the same as a normal function expression, which would be a visible change in function body handling.  So I agree now that the old structure, infix node with anonymous function at right, makes more sense.

I'll revert those soon, add a test so that `({f(){f();}});` is an error.

> we could simply skip over them, and not include them in the AST.

Well, yes, and tests have to be written to expect `class C {;}` to be printed as `class C {}`; besides that nothing really matters.

dgutov:
> add a test so that ({f(){f();}}); is an error.

Rather, an "undeclared variable or function" warning.

Also, please add a NEWS.md entry for this change.

XeCycle:
> Rather, an "undeclared variable or function" warning.

Ah yes, a warning.

> Also, please add a NEWS.md entry for this change.

You mean the rename?  Okay.

Btw mailing the assignment to FSF is quite expensive for me.  I may do that tomorrow.

dgutov:
> You mean the rename? Okay.

Yup.

> Btw mailing the assignment to FSF is quite expensive for me.

Is sending physical mail to US particularly expensive from China? I had to do this from Russia, and the cost was quite modest. It's not like you have to use a special order mail or something like that.

XeCycle:
I wiped away some intermediate changes in the refactor commit in this push, and it now looks a clean rename to me.

> Is sending physical mail to US particularly expensive in China?

DHL wants 150~200 RMB from me (as to how much this is, a dinner costs me 5~15 RMB).  It is not completely unaffordable to me, but I do consider this expensive.  Well, will send anyway.

dgutov:
> DHL wants 150~200 RMB from me (as to how much this is, a dinner costs me 5~15 RMB). It is not completely unaffordable to me, but I do consider this expensive. Well, will send anyway.

You could use the plain, postal mail instead, right? DHL will, of course, be faster, though.

XeCycle:
> You could use the plain, postal mail instead, right?

That sounds too time consuming.  If something changed on master while my mail is on the way it may require some unnecessary work from me.

But anyway, in what order would you like to merge my PRs?  I'd like to know where to base my future works on.  IMO the bind operator and this simple rename are basically unrelated, and can be merged in any order; also is the old async code.  But when adding async methods I'd like all three already there; anyway async methods is lowest priority on my side, since my current codebase contains no async methods.  I also plan to add ES7 property initializers, which I use now.

dgutov:
> If something changed on master while my mail is on the way it may require some unnecessary work from me.

Well, it's your call. I wouldn't expect merging the result to be too time-consuming.

> But anyway, in what order would you like to merge my PRs?

This one, probably. But it already exceeds the allowed limit without assignment, so I'll wait until it's completed first.

dgutov:
@XeCycle By the way, please avoid ending commit message summaries (the first line) with a period in the future.

Here's a good style guide to strive for: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

XeCycle:
Okay, messages edited.

XeCycle:
I have received the form signed by FSF, and rebased this to current master.  Guess I need to mail you that form?

dgutov:
Yes, please (you can see my email in the commit history). Maybe I won't have to ask that of contributors for much longer, though.

dgutov:
:ok_hand:

PR #278, opened 2015-10-29
`js2-print-json-path' added

redguardtoo:
place the cursor above key/value to print the json path which is also pushed into kill-ring

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -10878,6 +10878,67 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
         (js2-check-activation-name s (or token js2-NAME)))
     name))
 
+;;; Use AST to extract semantic information
+
+(defun js2-get-element-index-from-array-node (elem array-node &optional hardcoded-array-index)
+  "Get index of ELEM from ARRAY-NODE or 0 and return it as string."
+  (let ((idx 0) elems (rlt hardcoded-array-index))
+    (setq elems (js2-array-node-elems array-node))
+    (if (and elem (not hardcoded-array-index))
+        (setq rlt (catch 'nth-elt
+                    (dolist (x elems)
+                      ;; We know the ELEM does belong to ARRAY-NODE,
+                      (if (eq elem x) (throw 'nth-elt idx))
+                      (setq idx (1+ idx)))
+                    0)))
+    (format "[%s]" rlt)))
+
+(defun js2-print-json-path (&optional hardcoded-array-index)
+  "Print the path to the JSON value under point, and save it in the kill ring.
+If HARDCODED-ARRAY-INDEX provided, array index in JSON path is replaced with it."
+  (interactive "P")
+  (let (previous-node current-node
+        key-name
+        rlt)
+
+    ;; The `js2-node-at-point' starts scanning from AST root node.
+    ;; So there is no way to optimize it.
+    (setq current-node (js2-node-at-point))
+
+    (while (not (js2-ast-root-p current-node))
+      (cond
+       ;; JSON property node
+       ((js2-object-prop-node-p current-node)
+        (setq key-name (js2-prop-node-name (js2-object-prop-node-left current-node)))
+        (if rlt (setq rlt (concat "." key-name rlt))
+          (setq rlt (concat "." key-name))))
+
+       ;; Array node
+       ((or (js2-array-node-p current-node))
+        (setq rlt (concat (js2-get-element-index-from-array-node previous-node
+                                                                 current-node
+                                                                 hardcoded-array-index)
+                          rlt)))
+
+       ;; Other nodes are ignored
+       (t))
+
+      ;; current node is archived
+      (setq previous-node current-node)
+      ;; Get parent node and continue the loop
+      (setq current-node (js2-node-parent current-node)))
+
+    (cond
+     (rlt
+      ;; Clean the final result
+      (setq rlt (replace-regexp-in-string "^\\." "" rlt))
+      (kill-new rlt)
+      (message "%s => kill-ring" rlt))
+     (t
+      (message "No JSON path found!")))
+
+    rlt))
+
 ;;; Indentation support (bouncing)
 
 ;; In recent-enough Emacs, we reuse the indentation code from
--- a/tests/json-path.el
+++ b/tests/json-path.el
@@ -1,0 +1,64 @@
+;;; tests/json-path.el --- Test of using js2-mode AST to print JSON path.
+
+;; Copyright (C) 2015  Free Software Foundation, Inc.
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Code:
+
+(require 'ert)
+(require 'js2-mode)
+
+(ert-deftest js2-json-path-with-actual-array-index ()
+  (with-temp-buffer
+    (insert "var a = { hello: [1, 2, [1,3,3,4, { world: { hell: { yes: [1,2, 'test'] } } }]] };")
+    (js2-mode)
+    (goto-char 0)
+    (search-forward "test")
+    (should (string= (js2-print-json-path) "hello[2][4].world.hell.yes[2]"))))
+
+(ert-deftest js2-json-path-pure-arrays ()
+  (with-temp-buffer
+    (insert "var a = [5, 1, 4, [ 4, [1, 2, [1, 3.9, 4, [1, 2, 'test',3]]]], 9, 9];")
+    (js2-mode)
+    (goto-char 0)
+    (search-forward "test")
+    (should (string= (js2-print-json-path) "[3][1][2][3][2]"))))
+
+(ert-deftest js2-json-path-key-is-numeric ()
+  (with-temp-buffer
+    (insert "var b = {hello: {3 : {world: {2: 'test'}}}};")
+    (js2-mode)
+    (goto-char 0)
+    (search-forward "test")
+    (should (string= (js2-print-json-path) "hello.3.world.2"))))
+
+(ert-deftest js2-json-path-not-found ()
+  (with-temp-buffer
+    (insert "console.log('test');")
+    (js2-mode)
+    (goto-char 0)
+    (search-forward "test")
+    (should (eq (js2-print-json-path) nil))))
+
+(ert-deftest js2-json-path-with-array-index-hardcoded ()
+  (with-temp-buffer
+    (insert "var a = { hello: [1, 2, [1,3,3,4, { world: { hell: { yes: [1,2, 'test'] } } }]] };")
+    (js2-mode)
+    (goto-char 0)
+    (search-forward "test")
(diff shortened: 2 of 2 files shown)


js2-mode.el
```diff
@@ -10828,6 +10828,40 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
...
+  "Print the path to the JSON value under point, and save it in the kill ring."
+  (interactive)
+  (let (next-node node-type rlt key-name)
+    (setq next-node (js2-node-at-point))
+    ;; scanning from AST, no way to optimise `js2-node-at-point'
+    (while (and next-node (arrayp next-node) (> (length next-node) 5))
+      (setq node-type (aref next-node 0))
+      (cond
+       ;; json property node
+       ((eq node-type 'cl-struct-js2-object-prop-node)
```
dgutov: Why don't you use the predicate functions? Like `js2-object-prop-node-p`?

js2-mode.el
```diff
@@ -10828,6 +10828,40 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
...
     name))
 
+;;; Use AST to extract semantic information
+
+(defun js2-print-json-path ()
+  "Print the path to the JSON value under point, and save it in the kill ring."
+  (interactive)
+  (let (next-node node-type rlt key-name)
+    (setq next-node (js2-node-at-point))
+    ;; scanning from AST, no way to optimise `js2-node-at-point'
```
dgutov: Please capitalize and punctuate the comments.

dgutov:
Does this handle the case when the element in question is not the first item in an array?

Could you add a couple of tests?

redguardtoo:
@dgutov , thanks for the review. It's my intention to hardcode the array infix to zero. As a javascript developer, when I got the json sample like `{ key: [ 1, 2 , 3]`, my actual code will be like `key[i]`

So the final javascript code is actually always a variable name instead a number, use zero make replacing infix with my js variable name muc]h easier.

For example, assume evil-mode installed, for a json path "a[0].key[0].val[0]", I just need press "f0rif0rjf0rk"  and the path will be converted to "a[i].key[j].val[k]".

As you can see, "f0" is intentionally designed as duplicated key binding here, So if we use evil plugin called evil-space, we can decrease the key bindings to "f0ri rj rk".

I will fix other code style issue.

dgutov:
> It's my intention to hardcode the array infix to zero. 

This seems counter-intuitive, even if you know how to use the result.

Anyway, check this out: https://github.com/Sterlingg/json-snatcher

Unless you have a strong reason not to use that package, we should probably close this pull request.

redguardtoo:
Hi, @dgutov , thanks for the feedback. I strongly recommend my solution because I'm *too familiar with `json-snatcher` (one year experience in seven projects)*. It's simply NOT usable in real world projects.

json-snatcher is dependent on python and `json-mode`, it's too picky on the syntax.

dgutov:
> json-snatcher is dependent on python and json-mode,

Is it? I don't see it using Python anywhere. And it doesn't require json-mode, from what I can see.

> it's too picky on the syntax.

Did you file an issue?

I'd prefer to use json-snatcher, if possible, also because there are plans to add a similar command to the Emacs core. Then it will obsolete json-snatcher, and everyone will be happy. Eventually.

dgutov:
> I'm too familiar with json-snatcher (one year experience in seven projects)

That's a strong argument, of course, but I'd prefer to see a proper bug report there first.

redguardtoo:
I show you some real world examples, why jsons-print-path from json-snatcher is useless.

Mockup data is defined by backend developers or frontend lead developer so my only option is accept the existing standard. They usually do NOT follow strict json format. For example, use single quote instead of double quotes to wrap string, use C style comment or C++ style comment in mockup data. Some times they also use conditional statement in mockup data .

Then you check `jsons-parse` and `jsons-consume-token` defined in jsons-snatcher, it use regular expression to consume JSON token. To make jsons-snatcher to handle javascript syntax means *RE-INVENT of js2-mode*.  

I've done more than 20 enterprise projects around the world (Japan, China, US, France, Australia), All the mockup json data is like my sample. And I'm 99% sure other web project around the world use similar mockup.

The root cause is BUSINESS REQUIREMENT is CHANGING and TOO MANY CORNER CASES MUST BE COVERED in mockup. So people usually prefer use javacript instead json to create mockup.

Here is some sample mockup, 
real-world


btw, I double check `jsons-print-path` defined at https://github.com/redguardtoo/emacs.d/blob/master/site-lisp/json-snatcher/json-snatcher.el. Though it use variable name like python_str. It's not dependent on ptyhon.

dgutov:
@redguardtoo Fair enough.

Could you change the logic to output the actual indexes inside arrays, though? The current result is counter-intuitive, and when using Evil, I'm sure you could manipulate it just as easily either way.

redguardtoo:
Yes, I can. I will print actual index by default but there advanced user could manually turn on  "counter-intuitive" way.

redguardtoo:
I updated the code, now printing the actual array index by default.

Here is sample data:
```javascript
var a = {
  hello: [1, 2, [1,3,3,4, {
    world: {
      hell: {
        yes: [1,2, 'test']
      }
    }
  }]]
};
```

dgutov:
Thanks. Guess all I waiting for now is a couple of tests.

redguardtoo:
I've added unit test. Please check

js2-mode.el
```diff
@@ -218,6 +218,13 @@ are enabled, these will also be included.")
   "ES6 externs.  If `js2-include-browser-externs' is enabled and
 `js2-language-version' is sufficiently high, these will be included.")
 
+(defvar js2-print-json-path-with-actual-array-index t
```
dgutov: Why this variable?

js2-mode.el
```diff
@@ -218,6 +218,13 @@ are enabled, these will also be included.")
...
+  "Does the JSON path contain the actual array index?
+Or we just use zero as default value?")
+
+(defvar js2-json-path-max-depth 128
```
dgutov: Can't you do without this one?

js2-mode.el
```diff
@@ -10878,6 +10885,64 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
...
+;;; Use AST to extract semantic information
+
+(defun js2-get-element-index-from-array-node (elem array-node)
+  "Get index of ELEM from ARRAY-NODE or 0 and return it as string."
+  (let ((idx 0) elems (rlt 0))
+    (setq elems (js2-array-node-elems array-node))
+    (if (and elem js2-print-json-path-with-actual-array-index)
+        (setq rlt (catch 'nth-elt
+                    (dolist (x elems)
+                      (when (and (eq (aref elem 0) (aref x 0))
```
dgutov: I believe you should be able to implement this without using `aref` anywhere.

tests/json-path.el
```diff
@@ -0,0 +1,42 @@
+;;; tests/json-path.el --- Some tests for js2-mode.
+
+;; Copyright (C) 2009, 2011-2013  Free Software Foundation, Inc.
```
dgutov: Wrong copyright years.

tests/json-path.el
```diff
@@ -0,0 +1,42 @@
...
+
+;; Thik file is part of GNU Emacs.
```
dgutov: Typo.

tests/json-path.el
```diff
@@ -0,0 +1,42 @@
...
+
+(ert-deftest js2-json-path-with-actual-array-index ()
+  (with-temp-buffer
+    (insert "var a = { hello: [1, 2, [1,3,3,4, { world: { hell: { yes: [1,2, 'test'] } } }]] };")
+    (js2-mode)
+    (setq js2-print-json-path-with-actual-array-index t)
+    (goto-char 0)
+    (search-forward "test")
+    (should (string= (js2-print-json-path) "hello[2][4].world.hell.yes[2]"))))
+
```
dgutov: Extra empty line.

dgutov:
Thanks. But I've looked though the patch again, and it's not ready yet.

redguardtoo:
I updated code.

js2-mode.el
```diff
@@ -10878,6 +10878,69 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
...
+(defun js2-get-element-index-from-array-node (elem array-node &optional hardcoded-array-index)
+  "Get index of ELEM from ARRAY-NODE or 0 and return it as string."
+  (let ((idx 0) elems (rlt hardcoded-array-index))
+    (setq elems (js2-array-node-elems array-node))
+    (if (and elem (not hardcoded-array-index))
+        (setq rlt (catch 'nth-elt
+                    (dolist (x elems)
+                      ;; Is same node by comparing position
+                      ;; We already known the ELEM does belong to ARRAY-NODE
+                      ;; so avoid comparing node type for performance reason
```
dgutov: `elem` being an array is an implementation detail. It probably won't change, but the array indexes might. Use struct accessors.

js2-mode.el
```diff
@@ -10878,6 +10878,69 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
...
+                    0)))
+    (format "[%s]" rlt)))
+
+(defun js2-print-json-path (&optional hardcoded-array-index)
+  "Print the path to the JSON value under point, and save it in the kill ring.
+If HARDCODED-ARRAY-INDEX provided, actual array index in JSON path are replaced with it"
+  (interactive "P")
+  (let (previous-node current-node
+        key-name
+        (i 0) (json-path-max-depth 128)
```
dgutov: I still don't understand this variable. In which situation can the path become too long that you need to abort?

js2-mode.el
```diff
@@ -10878,6 +10878,69 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
...
+        rlt)
+
+    ;; The `js2-node-at-point' starts scanning from AST root node.
+    ;; So there is no way to optimize it.
+    (setq current-node (js2-node-at-point))
+
+    (while (and current-node
```
dgutov: `js2-node-at-point` always returns a non-nil value, and it's always a node. Why check that it's an array, or its length?

redguardtoo:
I updated the code.

js2-mode.el
```diff
@@ -410,6 +410,16 @@ which doesn't seem particularly useful, but Rhino permits it."
   :type 'boolean
   :group 'js2-mode)
 
+(defmacro js2-parent (current-node)
```
dgutov: You don't need these either.

Use `js2-node-parent` and `js2-node-pos`.

redguardtoo:
code updated.

js2-mode.el
```diff
@@ -10878,6 +10878,70 @@ And, if CHECK-ACTIVATION-P is non-nil, use the value of TOKEN."
...
+  "Get index of ELEM from ARRAY-NODE or 0 and return it as string."
+  (let ((idx 0) elems (rlt hardcoded-array-index))
+    (setq elems (js2-array-node-elems array-node))
+    (if (and elem (not hardcoded-array-index))
+        (setq rlt (catch 'nth-elt
+                    (dolist (x elems)
+                      ;; Is same node by comparing position?
+                      ;; Since we know the ELEM does belong to ARRAY-NODE,
+                      ;; we could check only offset field for performance reason.
+                      (if (= (js2-node-pos elem) (js2-node-pos x))
```
dgutov: Why not compare the nodes with `eq`?

redguardtoo: which one is more efficient? Is eq more efficient than two integer operations (one integer plus plus one integer compare)?

dgutov: Don't forget that you're doing two array element lookups first.

`(eq elem x)` is obviously more efficient.

redguardtoo:
code updated

tests/json-path.el
```diff
@@ -0,0 +1,64 @@
+;;; tests/json-path.el --- Test of using js2-mode AST to print JSON path.
+
+;; Copyright (C) 2009, 2011-2015  Free Software Foundation, Inc.
```
dgutov: Please fix the copyright line as well. It should be just 2015.

redguardtoo:
copyright updated

dgutov:
All right, thank you.

PR #279, opened 2015-10-30
Allow semicolons in a class body

XeCycle:
This is published ES6.  Do you want more tests?  I admit the test is a little too simple.


dgutov:
Looks fine. I'd ask for more tests if we find a bug in here somewhere.

PR #280, opened 2015-10-30
Add default parameters support to destructuring

### LINKED ISSUE #205: Default destructuring parameters
The following testcase breaks the coloration :

```js
function foo( bar, { a = { } } = { } ) {
    var { b = { } } = bar;
}
```

XeCycle:
Also published ES6.  This implementation happens to allow ES7 stage 1 property initializers https://github.com/jeffmo/es-class-static-properties-and-fields.

Perhaps this can fix #205.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
   with its related functions.  It already handles generator methods,
   and will in the future add support for async methods, so the old
   name will get more confusing.
+* Support for default parameters in destructuring.  It should work for both
+  objects and arrays, in both literals and function arguments.
 
 ## 20150909
 
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -1717,6 +1717,9 @@ the correct number of ARGS must be provided."
 (js2-msg "msg.destruct.assign.no.init"
          "Missing = in destructuring declaration")
 
+(js2-msg "msg.init.no.destruct"
+         "Binding initializer not in destructuring assignment")
+
 (js2-msg "msg.no.octal.strict"
          "Octal numbers prohibited in strict mode.")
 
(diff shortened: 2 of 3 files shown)


dgutov:
Please add a NEWS entry.

XeCycle:
Ah when writing NEWS I realized I forgot to test for function arguments.  Fixing that.

XeCycle:
The squashed commit added one test and one NEWS entry to the above two.

dgutov:
> This implementation happens to allow ES7 stage 1

Do you use it already?

Otherwise, if it's not too complicated, we generally prefer to disallow a (yet) invalid syntax.

XeCycle:
I do use it sparingly, but at this point it is not hard for me to maintain my local version, so no problem at my side.  Disallowing this should be simple, just check for `js2-is-in-destructuring` at related points.

Actually support for this is not accurate, in that the semicolon following the class element is not required.  If we are going to add that some day the solution must not be simply adding `class-p` to the check list of this part.

XeCycle:
As said in the commit message, tests for class bodies are intentionally left out.

dgutov:
Thanks!

RusAlex:
When it's expected to be supported ?

RusAlex:
Or how can I disable this kind of error and keep js2r mode working in a file ?

dgutov:
@RusAlex This should work already.

RusAlex:
I meant, can I use js2-mode with ES7 features like this ?

dgutov:
This pull request was about a ES6 feature: http://dmitrysoshnikov.com/ecmascript/es6-notes-default-values-of-parameters/#default-values-of-destructured-components

PR #282, opened 2015-10-31
Add js2-jsx-mode

### LINKED ISSUE #140: Highlighting & indenting embedded XML (E4X)?

jacksonrayhamilton:
Complements the [latest change to master](http://git.savannah.gnu.org/cgit/emacs.git/commit/?id=958da7ff63d1d647f45fd649654136da78529324). Partially resolves #140.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -60,6 +60,12 @@
 
 ;;   (add-to-list 'interpreter-mode-alist '("node" . js2-mode))
 
+;; Support for JSX is available via the derived mode `js2-jsx-mode'.  If you
+;; also want JSX support, use that mode instead:
+
+;;   (add-to-list 'auto-mode-alist '("\\.js\\'" . js2-jsx-mode))
+;;   (add-to-list 'interpreter-mode-alist '("node" . js2-jsx-mode))
+
 ;; To customize how it works:
 ;;   M-x customize-group RET js2-mode RET
 
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -95,6 +95,7 @@
       (require 'js2-old-indent)
     (defvaralias 'js2-basic-offset 'js-indent-level nil)
     (defalias 'js2-proper-indentation 'js--proper-indentation)
+    (defalias 'js2-jsx-indent-line 'js-jsx-indent-line)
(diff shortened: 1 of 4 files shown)


js2-mode.el
```diff
@@ -11373,6 +11373,17 @@ Selecting an error will jump it to the corresponding source-buffer error.
...
 
+;; We may eventually want js2-jsx-mode to derive from js-jsx-mode, but that'd be
+;; a bit more complicated and it doesn't net us much yet.
+;;;###autoload
+(define-derived-mode js2-jsx-mode js2-mode "JSX-IDE"
+  "Major mode for editing JSX code.
+Like `js-jsx-mode', which see.
+Indentation support requires Emacs 25+ (or the latest js.el)."
+  ;; Feature-detect JSX indentation support (in js-mode from Emacs 25+).
+  (if (fboundp 'js-jsx-indent-line)
```
dgutov: What if the user's Emacs is too old? Will it silently do nothing?

Seems like a recipe for confusion.

jacksonrayhamilton: We could print the message `Indentation support requires Emacs 25+ (or the latest js.el)`?

Or copy over the new indentation code to js2-old-indent.el?

dgutov: > Or copy over the new indentation code to js2-old-indent.el?

That's an option.

Or we just ask people to use `js2-minor-mode` with `js-jsx-mode`.

Or we put this mode in a separate file, and signal an error when it's loaded when the version of Emacs isn't high enough. Or simply signal this error from inside of `js2-jsx-mode`, and keep it in this file.

I think an error is better than a warning message.

jacksonrayhamilton: Decided to copy it over, as that should be easiest for users.

Bu

dgutov:
All right. Please add a NEWS entry as well.

dgutov:
Should we also have instructions how to use it (adding it to `auto-mode-alist`), in the top commentary?

dgutov:
Fantastic, thanks!

istinspring:
so now it's possible to use js2-mode for ReactJS?

PR #283, opened 2015-11-05
Adapt old async to master & async methods

XeCycle:
Conflicts resolved are listed in the message of merge commit.

I noticed branch `nil` tried to implement async methods, but quite incomplete.  So I wrote async methods myself.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,15 +4,16 @@
 
 * Object properties are highlighted using a different face:
   `js2-object-property`, which has no color by default.
-* `js2-getter-setter-node` is renamed to `js2-method-node`, together
-  with its related functions.  It already handles generator methods,
-  and will in the future add support for async methods, so the old
-  name will get more confusing.
+* `js2-getter-setter-node` is renamed to `js2-method-node`, together with
+  its related functions.  It already handles generator methods, and we
+  added support for async methods (see below), so the old name will get
+  more confusing.
 * Support for default parameters in destructuring.  It should work for both
   objects and arrays, in both literals and function arguments.
 * New mode: `js2-jsx-mode`, deriving from `js2-mode`.  Supports indentation of
   JSXElement expressions wrapped within parentheses or as function arguments.
   Indentation is customizable via `sgml-attribute-offset`.
+* Experimental support for ES7 stage 3 async/await.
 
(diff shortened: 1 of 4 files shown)


XeCycle:
Sorry I forgot to pass async-p to parse-function-expr.  Added a bugfix commit; if you think all these looks okay I will squash it into "parse async methods".

Btw do we have a fixed policy on what proposals are ok to implement?  Decide on stage (e.g. >=3) or popularity, or simply reject all?

dgutov:
> Sorry I forgot to pass async-p to parse-function-expr. Added a bugfix commit; if you think all these looks okay I will squash it into "parse async methods".

Looks good. I think you might want to squash only the last two commits together ("bugfix" with the previous one), or just keep them all as they are.

> Decide on stage (e.g. >=3) or popularity, or simply reject all?

>=3 sounds good. Rejecting all would be inconsistent, since it seems we're going to accept the "Async Functions" one. :) It looks fairly popular.

If you're motivated enough to experiment with more riskier proposals, we should probably be able to disable them by default. For instance, by updating the `js2-language-version` variable semantics for the current language versions, and using it consistently. So it would default to ES6, add could additionally be set to different stages of ES7, too.

XeCycle:
Squashed last two.

> updating the js2-language-version variable semantics for the current language versions, and using it consistently. So it would default to ES6, add could additionally be set to different stages of ES7, too.

So we set some version numbers for different stages of ES7?  Pls decide, and I'll make this conditional later.  Playing with old commits are nasty...  they may conflict at any time.  Let's merge first and add that later.

Btw the build failure seems to be failing to install emacs.

dgutov:
> Squashed last two.

Thanks! Merging.

> So we set some version numbers for different stages of ES7? Pls decide, and I'll make this conditional later.

Yes, if you want to work on implementing features.

> Playing with old commits are nasty... they may conflict at any time. Let's merge first and add that later.

Sure. I'm sure we can live with unconditional support for async functions for a while.

> Btw the build failure seems to be failing to install emacs.

Indeed, I had to restart it.

dgutov:
Before I merge, though, please add a NEWS entry (and update the `js2-getter-setter-node` news item).

PR #285, opened 2015-11-05
Fix parsing of param with destructure and default

dgreensp:
`function foo(bar, {a = {}} = {}) {/**/}`

Fixes the original test case in #205.

What I did (see if it looks right):
* Pulled `(push param params)` out of the `cond`, since it happens no matter what
* Pulled `;; default parameter value` handling out of the `cond`, so that it happens even for the destructuring case
* Added the check `(not rest-param-at)` to `;; default parameter value` handling so that it doesn't trigger in case of a rest param

I added a test but I did not run the tests.

The existing NEWS item should cover this change.


dgutov:
Thanks!

PR #287, opened 2015-11-06
Fix semicolons and scoping in exported decls

### LINKED ISSUE #215: Declaration warning after named function export
Consider the following script:

```js
export function foo () {}

foo();
```

`js2-mode` marks the `foo()` statement with the following warning: "Undeclared variable or function 'foo'".

It seems to me from the [ES6 spec](http://people.mozilla.org/~jorendorff/es6-draft.html#table-42), that the function should also be declared locally. This is also how the [ES6 Module Transpiler](http://esnext.github.io/es6-module-transpiler/) works.

dgreensp:
In the following code, no semicolons should be considered missing,
and all of A,B,C,D should be considered declared.

```
export function A() {}
export class B {}
export default function C() {}
export default class D {}

var x = [A, B, C, D];
```

This should fix #215.

The approach here of parsing functions and classes as statements seems to work.

See also;
* [`export` language grammar](http://www.ecma-international.org/ecma-262/6.0/#sec-exports)
* [Discussion of semicolons over at eslint](https://github.com/eslint/eslint/issues/2194)

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -8842,20 +8842,30 @@ invalid export statements."
               (setq from-clause (js2-parse-from-clause)))
           (js2-unget-token))))
      ((js2-match-token js2-DEFAULT)
-      (setq default (js2-parse-expr)))
+      (setq default (cond ((js2-match-token js2-CLASS)
+                           (js2-parse-class-stmt))
+                          ((js2-match-token js2-FUNCTION)
+                           (js2-parse-function-stmt))
+                          (t (js2-parse-expr)))))
      ((or (js2-match-token js2-VAR) (js2-match-token js2-CONST) (js2-match-token js2-LET))
       (setq declaration (js2-parse-variables (js2-current-token-type) (js2-current-token-beg))))
+     ((js2-match-token js2-CLASS)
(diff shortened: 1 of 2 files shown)


dgreensp:
It looks like exported `class` and `function` declarations pretty-print with some extra newlines after this change.  Not sure if that's worth doing something about or not.

dgutov:
Thanks for working on this. I think the printing of the newline before semicolons looks suspicious, though. Do you know why that happens?

dgreensp:
It's pretty simple; the parse of `export function A() {}` goes from "export function-expression" to "export function-statement", and the pretty-printer specifically inserts those two extra newlines when it pretty-prints a function declaration vs. function expression; see https://github.com/mooz/js2-mode/blob/e0972fbc11975bd1129ee42a422a3092d469c132/js2-mode.el#L3328 .  I don't think that behavior is super principled.

dgutov:
I see. And we need to use `js2-parse-function-stmt`, because function expressions don't declare their names in the current scope.

Fixing the printing doesn't seem too hard, though. See the follow-up commit.

XeCycle:
This did not cover async functions.  Working on it.

Issue #248, opened 2015-07-26
spread in object literal results in error


gaydenko:
Hi!
```js
const obj = {};

const obj2 = {
    ...obj,
  plus: 'plus'
};
```
For such spread operator I get `invalid property id` error on `...obj` (followed with other derived errors). Is it possible to workaround the issue?

dgutov:
There's an attempt to implement support for this in #245, but it seems to have stalled.

gaydenko:
Thanks for the clarification. Must I keep the issue opened? Or - is PR sufficient?

dgutov:
Yes, let's keep it open.

nikki93:
hey any updates on object spread operator? it's pretty useful in programming with react libraries!

dgreensp:
What should the AST look like for `var obj = {a:1, b:2, ...more, ...foo(1, 2, 3)}`?  It seems clear that `...foo(1,2,3)` should become an `object-prop` node whose `right` (not `left`) is `foo(1,2,3)`, with the `left` being some sentinel value that won't trip up any existing code that iterates over `object` nodes too much.  What should that value be?  Something based on the `TRIPLEDOT` token?

dgreensp:
Hmm, or maybe `...foo(1,2,3)` doesn't have to become an `object-prop` after all.  I was assuming that all `elems` of an `object-node` should be `object-prop` nodes, but it looks like e.g. methods are already a completely different struct, and it works just fine.  So maybe a `unary-node`.

dgutov:
> it looks like e.g. methods are already a completely different struct, and it works just fine

`js2-method-node` inherits from `js2-infix-node`, though.

dgreensp:
I signed with FSF.

I haven't had a chance to revise my PR, but I am on board with using an `infix` node for less risk of breakage.  When I come back to it, I'll figure out how to make an `infix` node and what `:left` and `:type` should be.

jeffbski:
+1

kamilwaheed:
+1

Really looking forward to this getting addressed.

damonmcminn:
+1

azer:
+1

miguel-guedes:
+1

edheil:
+1

cauchycai:
+1

dgutov:
This should work now. Thanks to @dgreensp and the testers.

PR #289, opened 2015-11-08
Fix array destructuring including triple-dot

dgreensp:
This change enables proper parsing of:
* `let [a, b, ...c] = d;`
* `let [[x,y] = [1,2]] = z;`

The first case is simply the "rest" part of "array rest/spread," which
lacked the appropriate cases in `js2-parse-array-literal` and
`js2-define-destruct-symbols`.  There's also new logic to make sure that
in the destructuring (rest) case, the rest element comes last (while
spread can occur anywhere).

The removal of the `;; destructuring bind` case of
`js2-parse-array-literal` may require some explanation.  It was doing
more harm than good.  For example, it assumed that only identifiers
have defaults.  Also, calling `js2-parse-destruct-primary-expr` is
unnecessary, since `js2-is-in-destructuring` is already true.  Rather
than add the triple-dot logic in two places, it was simpler to stop
distinguishing between destructuring and non-destructuring at this
level, because the grammars are extremely similar.  In addition, wacky
"vars" as in `let [f()] = x` throw an error anyway when they hit
`js2-define-destruct-symbols`.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -8079,7 +8079,11 @@ declared; probably to check them for errors."
      ((js2-array-node-p node)
       (dolist (elem (js2-array-node-elems node))
         (when elem
-          (if (js2-infix-node-p elem) (setq elem (js2-infix-node-left elem)))
+          (setq elem (cond ((js2-infix-node-p elem) ;; default (=)
+                            (js2-infix-node-left elem))
+                           ((js2-unary-node-p elem) ;; rest (...)
+                            (js2-unary-node-operand elem))
+                           (t elem)))
           (push (js2-define-destruct-symbols
                  elem decl-type face ignore-not-in-block)
                 name-nodes)))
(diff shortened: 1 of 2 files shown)


dgreensp:
I'm working up to fixing #248.

dgutov:
Thank you, David.

Please clarify: do you have copyright assignment to FSF signed?

dgreensp:
Not to my memory.  Where do I sign?

dgutov:
You can follow these instructions: http://git.savannah.gnu.org/cgit/gnulib.git/tree/doc/Copyright/request-assign.future

Put "Emacs" as the answer to the first question.

dgutov:
@dgreensp Any last-minute fixes, or should I just merge this now?

astephenb:
Hi there! I was wondering when this might get merged? Most of my team is using js2 mode now and would love to see the spread operator supported!

dgutov:
@astephenb Have you tried this patch out? Does it work well?

Note that this one doesn't exactly add support for the spread operator.

astephenb:
@dgutov Sorry I have not. I just saw the last comment on this was a few months ago and figured it was good to go! I will try to test out the patch.

dgutov:
All right, this looks good. Thanks, David!

PR #290, opened 2015-11-09
Also parse async function statement in export node

XeCycle:
AsyncFunctionDeclaration is added as a
HoistableDeclaration (http://tc39.github.io/ecmascript-asyncawait/#HoistableDeclaration),
which shall be covered in productions "export Declaration" and "export
default HoistableDeclaration".


tests/parser.el
```diff
@@ -799,6 +806,12 @@ the test."
 (js2-deftest-parse parse-export-generator-declaration "export default function* one() {\n}")
 (js2-deftest-parse parse-export-assignment-expression "export default a = b;")
 
+(js2-deftest-parse parse-export-async-function-allow-await
+  "export async function f() {\n  await f();\n}\n;")
```
XeCycle: Note here; the trailing `\n;` also appears in `export function f() {}`.

dgutov: Will you add a corresponding fix to `js2-print-export-node`?

XeCycle: Okay, will make that fix commit before this one.

dgutov:
Thanks!

PR #293, opened 2015-11-11
Implement object literal rest/spread (...)

dgreensp:
As discussed in #245, this syntax is not part of ES6/2015, but it is
a stage 2 proposal, meaning the committee expects it to be included
in the standard.  It's already seeing heavy use by e.g. the React
community.

This change causes `...expr` in an object literal to be parsed as a
unary-node element of an object-node, and modifies places where an
element of an object-node is expected to be an infix-node.  Note that
only weak typing assumptions are made already about object-node elems.

In addition, this change implements proper name extraction in
object literal destructuring.  The cases are:

1. `let {a} = x;`
2. `let {a: b} = x;` (should define `b`)
3. `let {a = b} = x;` (should define `a`)
4. `let {...a} = x;` (should define `a`)

This change fixes case (2), which was broken, and adds case (4).

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -8079,7 +8079,11 @@ declared; probably to check them for errors."
      ((js2-array-node-p node)
       (dolist (elem (js2-array-node-elems node))
         (when elem
-          (if (js2-infix-node-p elem) (setq elem (js2-infix-node-left elem)))
+          (setq elem (cond ((js2-infix-node-p elem) ;; default (=)
+                            (js2-infix-node-left elem))
+                           ((js2-unary-node-p elem) ;; rest (...)
+                            (js2-unary-node-operand elem))
+                           (t elem)))
           (push (js2-define-destruct-symbols
                  elem decl-type face ignore-not-in-block)
                 name-nodes)))
@@ -10449,19 +10453,13 @@ array-literals, array comprehensions and regular expressions."
 
 (defun js2-parse-array-literal (pos)
   (let ((after-lb-or-comma t)
-        after-comma tt elems pn
+        after-comma tt elems pn was-rest
         (continue t))
     (unless js2-is-in-destructuring
       (js2-push-scope (make-js2-scope))) ; for the legacy array comp
     (while continue
       (setq tt (js2-get-token))
       (cond
-       ;; comma
-       ((= tt js2-COMMA)
-        (setq after-comma (js2-current-token-end))
-        (if (not after-lb-or-comma)
-            (setq after-lb-or-comma t)
-          (push nil elems)))
        ;; end of array
        ((or (= tt js2-RB)
             (= tt js2-EOF))  ; prevent infinite loop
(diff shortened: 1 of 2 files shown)


dgreensp:
I'll fix this PR before merge to not include the other PR.  Just wanted to get it in front of you.

I'm working on the FSF copyright assignment.

Thanks!

dgutov:
> I'll fix this PR before merge to not include the other PR.

Not a problem, I'll just merge the other one first.

Thank you.

dgutov:
> This change causes ...expr in an object literal to be parsed as a unary-node element of an object-node

That might be the best we can do, even though it might break some third party code that relies on object-node elements being infix-nodes. But better ideas welcome.

dgreensp:
Maybe it could/should be an infix (though we get AST printing for free this way).  I'll come back to it tonight or tomorrow and give it another thought.

XeCycle:
IMO you may try inheriting from infix node, and override the print function.  A simple binding / shorthand property is an infix node, e.g. `{a}`, fwiw.

I feel we'd better assign version numbers for ES7 features, preferably several numbers for different stages, and make these features conditional.

dgutov:
> I feel we'd better assign version numbers for ES7 features, preferably several numbers for different stages, and make these features conditional.

Yes. Implementing this feature, though, should be fairly orthogonal to the pull request in question.

XeCycle:
> should be fairly orthogonal to the pull request in question

If version numbers are decided, this pr may need changes to be conditional on version number, right?

dgutov:
> this pr may need changes to be conditional on version number, right?

Of course. But I think they'll be minor enough.

XeCycle:
> they'll be minor enough.

Okay, agreed.

jeffbski:
+1

iensu:
Hi guys!

What is the status of this PR? Still waiting for the ES7 feature version numbers?

dgreensp:
I haven't had time to work on it lately.

iensu:
@dgreensp Ok I understand :) 
Would be really awesome if it made it into js2-mode though. 
I just patched my local version with this PR, and it seems to be working great so far. Please, tell me if there's anything I assist with.

buob:
I second this plus, this would be great to have!  A bunch of us here at our company would love to get this in, as we're using spread operators everywhere. :grin:  I also second the offer for help if there's anything.

dgutov:
@buob Then please try out this patch and report back.

zakandrewking:
This patch is working great for me. The only potentially odd thing is the indentation when the spread is on it's own line. It looks like this now:

```
const data = {
    ...old_data,
  val: 10
}
```

I might have expected this:

```
const data = {
  ...old_data,
  val: 10
}
```

But still, now my js2-mode works for these new tricks in ES7/React.

dgutov:
@zakandrewking Thanks for checking.

The indentation problem was reported in http://debbugs.gnu.org/23492, I'll port it to `js2-old-indent` shortly.

dgutov:
Actually, never mind. It's already there.

PR #297, opened 2015-11-30
Recognize for-of and for-in loops with const

Christoph-D:
Declarations in for-of and for-in loops may also be "const".  See
http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-bindinginstantiation


dgutov:
Thanks!

Apparently, this is the relevant Mozilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1094995

PR #301, opened 2015-12-11
Fix errors shown when async module is used

### LINKED ISSUE #300: Usage of async library leads to warnings
When I use a method from [caolan/async](https://github.com/caolan/async), js2-mode highlights `async` as a keyword and shows a warning under the first letter, that the code has no side effects.

It would be nice if js2-mode could either identify this and treat `async.foo` calls as normal method calls, or at least have a variable where it can be set.

alanpearce:
Fixes #300

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -7749,7 +7749,8 @@ string is NAME.  Returns nil and keeps current token otherwise."
 
 (defun js2-match-async-arrow-function ()
   (when (and (js2-contextual-kwd-p (js2-current-token) "async")
-             (/= (js2-peek-token) js2-FUNCTION))
+             (/= (js2-peek-token) js2-FUNCTION)
+             (/= (js2-peek-token) js2-DOT))
     (js2-record-face 'font-lock-keyword-face)
     (js2-get-token)
     t))


XeCycle:
Hmm, good catch, but this of course is not enough.  It would need to look ahead, skip over a name or a paren expr, and check the next token is an arrow.

alanpearce:
Ah, okay then. I've never used the new `async` keyword, so perhaps it's best I leave this to someone else who understands both sides.

dgutov:
> It would need to look ahead, skip over a name or a paren expr, and check the next token is an arrow.

What will it be otherwise? A plain identifier?

XeCycle:
> What will it be otherwise? A plain identifier?

I think so.  The proposal defined it similar to the original arrow function, by covering the special case in `CallExpression` (http://tc39.github.io/ecmascript-asyncawait/#CallExpression); so maybe we can mimic what `js2-parse-assign-expr` does, save a tokenizer state at beginning in `js2-parse-member-expr`.

dgutov:
> maybe we can mimic what js2-parse-assign-expr does, save a tokenizer state at beginning in js2-parse-member-expr

I suppose so. Unless anyone else has a better idea.

dgutov:
Implemented differently in #304.

PR #304, opened 2016-01-03
Fix void async await

jacksonrayhamilton:
Fixes some parsing errors with unary expressions.

A little while ago I observed the issue described in #300, and finally I was spurred to fix it. I discovered `js2-match-async-arrow-function` caused an off-by-one error, so I prevented that.

I also noticed that `await` parsing was too eager, so I added an extra guard there.

I also noticed that `void` was not being printed correctly with `js2-print-tree`, so I fixed that too.


dgutov:
This is a minor nit, but could you please also remove the periods at the end of commit messages?

dgutov:
Thanks a lot!

PR #308, opened 2016-01-24
Add support for JSDoc @callback, @func and @method tags

### LINKED ISSUE #307: Feature request: Add support for jsdoc @callback tag
Currently, the [`@callback`](http://usejsdoc.org/tags-callback.html) tag isn't highlighted. It's just like the [`@function`](http://usejsdoc.org/tags-function.html) tag, so support can be easily added.

While we're at it, the two aliases `@func` and `@method` of `@function` are also unhandled. The following quick patch will fix all three:

```diff
diff --git a/js2-mode.el b/js2-mode.el
index aa1d60d..ac15af5 100644
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -6906,15 +6906,18 @@ of a simple name.  Called before EXPR has a parent node."
            '("alias"
              "augments"
              "borrows"
+             "callback"
              "bug"
              "base"
              "config"
              "default"
              "define"
              "exception"
+             "func"
              "function"
              "member"
              "memberOf"
+             "method"
              "name"
              "namespace"
              "since"
@@ -6931,6 +6934,7 @@ of a simple name.  Called before EXPR has a parent node."
           (regexp-opt
            '("addon"
              "author"
+             "callback"
              "class"
              "const"
              "constant"
@@ -6945,6 +6949,7 @@ of a simple name.  Called before EXPR has a parent node."
              "export"
              "fileoverview"
              "final"
+             "func"
              "function"
              "hidden"
              "ignore"
@@ -6953,6 +6958,7 @@ of a simple name.  Called before EXPR has a parent node."
              "inner"
              "interface"
              "license"
+             "method"
              "noalias"
              "noshadow"
              "notypecheck"
```

zmwangx:
See http://usejsdoc.org/tags-callback.html for documentation of the
@callback tag; see http://usejsdoc.org/tags-function.html for
documentation of the @func and @method tags (both are synonyms of the
@function tag).

Closes #307.


dgutov:
Thanks.

Could you give an example of what doesn't work if you don't add these values to `js2-jsdoc-empty-tag-regexp`?

zmwangx:
Sorry, I didn't realize the `<namepath>` argument to `@callback` is mandatory. I have removed `callback` from `js2-jsdoc-empty-tag-regexp`.

Re `@func` and `@method`: like `@function`, the `<FunctionName` argument is optional, so they should be in the empty tag group, or something like

```
/**
 * @method
 */
```

won't work.

dgutov:
Great, thanks for the update. :+1:

PR #326, opened 2016-04-08
Fix a byte compilation warning on Emacs 25

phst:


dgutov:
That might be a bit premature, but if you insist, please rebase.

phst:
Rebased

Issue #295, opened 2015-11-17
`export default function() {}` shouldn't require a name


dgreensp:
`export default function() {}` is currently an error, though it should be legal.  `export function () {}` is correctly illegal.

This comes out of my own PR #287, which parses exported functions as statements/declarations so that the function name is considered to be in scope of the surrounding code.  For the `export default` case, we just have to also support the case where there's no name (and maybe parse it as an expression in that case instead of a statement).

dgreensp:
Alternatively, we allow a function declaration to be anonymous, but only inside `export default`, somehow.  This is what the spec does (see note at the bottom of http://www.ecma-international.org/ecma-262/6.0/#sec-function-definitions-runtime-semantics-instantiatefunctionobject).

The same problem applies to `export default class {}`.

dgutov:
Would there be some practical difference between a anonymous function expression and an anonymous declaration, in this case?

dgreensp:
I don't think so, no.

dgutov:
Then option one probably makes more sense.

PR #327, opened 2016-04-26
Colorize contextual keyword as in import & export

XeCycle:
No idea how to test this.  Required?

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -8632,50 +8632,47 @@ imports or a namespace import that follows it.
   "Parse a namespace import expression such as  '* as bar'.
 The current token must be js2-MUL."
   (let ((beg (js2-current-token-beg)))
-    (when (js2-must-match js2-NAME "msg.syntax")
-        (if (equal "as" (js2-current-token-string))
-            (when (js2-must-match-prop-name "msg.syntax")
-              (let ((node (make-js2-namespace-import-node
-                           :pos beg
-                           :len (- (js2-current-token-end) beg)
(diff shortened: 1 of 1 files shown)


XeCycle:
Amended, added `from`.

js2-mode.el
```diff
@@ -8870,12 +8865,10 @@ invalid export statements."
...
         (dolist (export exports-list)
           (push export children)))
-      (when (js2-match-token js2-NAME)
-        (if (equal "from" (js2-current-token-string))
-            (progn
-              (js2-unget-token)
-              (setq from-clause (js2-parse-from-clause)))
-          (js2-unget-token))))
+      (when (js2-match-contextual-kwd "from")
+        (progn
```
dgutov: Don't need `progn` inside `when`.

dgutov:
Good idea, thanks.

Testing this isn't hard: you would call `get-text-property` at a given position. Maybe find that position with `(search-backward "as ")`. But we haven't been writing tests like that so far, so no, it's not required.

XeCycle:
Fixed that progn, still no tests.  Btw another commit is here https://github.com/XeCycle/js2-mode/commit/292d1e5a82279e6dc69b6c410b02c9b08bdd07be, how do you like this one?

dgutov:
Thanks!

>  Btw another commit is here XeCycle@292d1e5, how do you like this one?

Looks okay. Does it fix #295?

XeCycle:
> Does it fix #295?

No I don't think them related.  Will open a PR for that anyway.

PR #332, opened 2016-05-02
Allow space before global declaration

bolivier:
I added the allowance for an optional space in a global declaration. 

Eslint syntax for global declarations differs slightly from jslint, and this change makes that syntax considered valid.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -7264,7 +7264,7 @@ are ignored."
            when (and (eq 'block (js2-comment-node-format node))
                      (save-excursion
                        (goto-char (js2-node-abs-pos node))
-                       (looking-at "/\\*global ")))
+                       (looking-at "/\\* *global ")))
            append (js2-get-jslint-globals-in
                    (match-end 0)
                    (js2-node-abs-end node))))
--- a/tests/externs.el
+++ b/tests/externs.el
@@ -42,6 +42,13 @@
     (should (equal (js2-get-jslint-globals)
                    '("quux" "tee" "$")))))
 
+(ert-deftest js2-finds-jslint-globals-with-space ()
+  (with-temp-buffer
+    (insert "/* global foo, bar:false, baz:true")
+    (js2-mode)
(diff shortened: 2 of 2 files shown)


js2-mode.el
```diff
@@ -7264,7 +7264,7 @@ are ignored."
            when (and (eq 'block (js2-comment-node-format node))
                      (save-excursion
                        (goto-char (js2-node-abs-pos node))
-                       (looking-at "/\\*global ")))
+                       (looking-at "/\\*[ ]?global ")))
```
dgutov: There's no need for the square brackets here. Maybe it should be ` *`, though. Does ESLint accept multiple spaces before `global`?

bolivier: Eslint does work with multiple spaces.  I think `" +` is better because characters other than space probably shouldn't be accepted.  `/* x global foobar */` shouldn't work, in my opinion.

js2-mode.el
```diff
@@ -7264,7 +7264,7 @@ are ignored."
            when (and (eq 'block (js2-comment-node-format node))
                      (save-excursion
                        (goto-char (js2-node-abs-pos node))
-                       (looking-at "/\\*global ")))
+                       (looking-at "/\\* ?+global ")))
```
bolivier: This should cover valid cases.

dgutov: Why not `" *"` instead of `" ?+"`?

bolivier: `" *"` would match any characters, including other words, in front of `global`.  If you have non space characters before `global` it shouldn't be considered valid.

dgutov: > " *" would match any characters, including other words, in front of global.

Not at all. Just try it.

The difference between `+` and `*` is `{1,}` vs `{0,}`.

bolivier: I was using regexp builder in Emacs and I was nearly certain that it matched things it shouldn't. 

I'll double check.

dgutov:
:thumbsup:

PR #333, opened 2016-05-04
Change warning message to not conflict with spec

daniellandau:
The ECMAScript spec says
(http://www.ecma-international.org/ecma-262/5.1/#sec-10.5) that
declaring an already declared variable is silently ignored, so claiming
it's a TypeError is not correct. It should still be highlighted as a
warning because it's not something that any program should do and with
let redeclaration actually is a TypeError.


dgutov:
Sounds good, thanks.

PR #342, opened 2016-06-10
Add js2-mode-assume-strict custom variable.

### LINKED ISSUE #341: Allow setting strict mode by default
Currently js2-in-use-strict-directive is a deflocal that is reset to nil in js2-do-parse.  It would be great to add a defcustom to allow defaulting it to t instead (i.e. for ES6 modules, which are strict by default, or else for compiled JS that are parsed in strict mode based on e.g. a command-line flag).

shicks:
This allows all files to be parsed as if 'use strict' were on top of every file, for the case of JS passed through compilers that assume strict input.

Closes #341.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -277,6 +277,11 @@ end of the line, otherwise, at the beginning of the next line."
   :type 'boolean
   :group 'js2-mode)
 
+(defcustom js2-mode-assume-strict nil
+  "Non-nil to start files in strict mode automatically."
+  :type 'boolean
+  :group 'js2-mode)
+
 (defcustom js2-mode-show-strict-warnings t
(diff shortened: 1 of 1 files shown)


dgutov:
Looks good, but please remove the period at the end of the message summary.

shicks:
I can do that, but all the other ones nearby have periods.  Are you sure you want it removed?

dgutov:
You mean other sentences? That's entirely fine.

See http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html.

shicks:
I thought you meant the emacs defcustom message.  Fixed.

dgutov:
Thanks, and sorry about the ambiguity.

PR #345, opened 2016-06-23
Fix Highlighting function name of form {"quoted name" () {...}}

jacott:
This patch will make "quoted name" use font-lock-function-name-face just
like {"quoted name": function () {...}} does.

I tried to write a test for this but could find no existing tests for checking
face settings.


dgutov:
Probably no need for a test here, but please move the comment so it doesn't cross the 80th column.

jacott:
I've updated the comment above. Note the test comment is not in the actual commit

dgutov:
Sorry, what did you update? I meant the comment at the end of the line.

jacott:
Oh okay; I thought you meant the commit comment :/

PR #347, opened 2016-06-25
Makefile: allow overriding $EMACS

dunn:


dgutov:
Are you sure overriding doesn't work without this change? Because it does here.

dunn:
Something like `EMACS=$(brew --prefix)/bin/emacs make test` wasn't working for me before this change.

dgutov:
The customary usage is `make test EMACS=$(brew --prefix)/bin/emacs`.

dunn:
Oh, right, I was using it as an env var instead of a make variable.

I'll close this, then.  Thanks!

PR #353, opened 2016-07-07
js2-jsx-mode docstring fix

kaushalmodi:
Reference: http://lists.gnu.org/archive/html/emacs-devel/2016-07/msg00297.html

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -11620,7 +11620,7 @@ variables (`sgml-basic-offset' et al) locally, like so:
 
   (defun set-jsx-indentation ()
     (setq-local sgml-basic-offset js2-basic-offset))
-  (add-hook 'js2-jsx-mode-hook #'set-jsx-indentation)"
+  (add-hook \\='js2-jsx-mode-hook #\\='set-jsx-indentation)"
   (set (make-local-variable 'indent-line-function) #'js2-jsx-indent-line))
 
 (defun js2-mode-exit ()


dgutov:
Please attribute the commit to the right author.

kaushalmodi:
Apologies, there was no intention to steal the attribute .. I proposed the exact same fix on the emacs-devel thread in parallel here: http://thread.gmane.org/gmane.emacs.devel/205313/focus=205321

*I am figuring out how to change the author in magit now ..*

Please do so from your end if possible for now.

dgutov:
> Please do so from your end if possible for now.

I can't do that with the "merge pull request" button, and it would be a poor form to edit the commit afterwards.

> I proposed the exact same fix on the emacs-devel thread in parallel here

So you didn't use Paul's patch?

kaushalmodi:
> So you didn't use Paul's patch?

No. just that the fix was pretty straight forward and I did the exact same thing on my end... simply adding `\\=` before the single quotes.

dgutov:
Cool, thanks.

kaushalmodi:
Thanks. But this brings up an interesting question .. how to amend the commit author using magit if need be for the next time :)

Please share if you know how to do so.

dgutov:
> how to amend the commit author using magit

I dunno. With VC, it's done using `Amend: true` and `Author: ...` headers in the commit message buffer.

kaushalmodi:
OK, for anyone else reading this, here is how to change author in magit in the last commited hunk:

- `c` 
- `=A`
- *type the new author name email in minibuffer* `"Foo Baar <foo@bar.com>"`
- `a`
- `C-c C-c`

PR #358, opened 2016-08-04
Add support for /*jslint*/ declarations

### LINKED ISSUE #237: Parse /* jslint */ declarations
js2-mode should be able to parse `/* jslint */` declarations. This is especially useful for e.g. Node.js code, where you'd want to have `/* jslint node */`.

Looking at `M-x customize-group RET js2-mode`, it appears that `js2-mode` already has the capability to do this. It just needs to be hooked up to some parsing logic.

JulianKniephoff:
Closes #237

I based this heavily on (and reused parts of) the mechanics to support `/*global*/` declarations. The lists of externs for the different declarations is lifted directly from the [JSLint documentation](http://www.jslint.com/help.html).

### DIFF
--- a/tests/externs.el
+++ b/tests/externs.el
@@ -59,3 +59,68 @@
 ;;;TODO
 ;; ensure that any symbols bound with the import syntax are added to the extern list
 ;; ensure that any symbols bound with the export syntax exist in the file scope
+
+;; Test `/*jslint*/` declarations
+
+(ert-deftest js2-finds-jslint-declaration-externs ()
+  (with-temp-buffer
+    (insert "/*jslint browser: true, node: true*/")
+    (js2-mode)
+    (let ((found-externs (js2-get-jslint-declaration-externs)))
+      (should (and (member "navigator" found-externs)
+                   (member "__filename" found-externs))))))
+
+(ert-deftest js2-finds-jslint-declaration-externs-without-property-value ()
+  (with-temp-buffer
+    (insert "/*jslint devel, es6: true, couch")
+    (js2-mode)
+    (let ((found-externs (js2-get-jslint-declaration-externs)))
+      (should (and (member "prompt" found-externs)
+                   (member "Promise" found-externs)
+                   (member "require" found-externs))))))
+
+(ert-deftest js2-no-jslint-declaration-externs-without-keyword ()
+  (with-temp-buffer
+    (insert "/* browser: true, node: true*/")
+    (js2-mode)
+    (should (null (js2-get-jslint-declaration-externs)))))
+
+(ert-deftest js2-no-jslint-declaration-externs-for-nonexistent-env ()
+  (with-temp-buffer
+    (insert "/*jslint nonexistent: true*/")
+    (js2-mode)
+    (should (null (js2-get-jslint-declaration-externs)))))
+
+(ert-deftest js2-finding-jslint-declaration-externs-ignores-nonexistent ()
+  (with-temp-buffer
+    (insert "/*jslint es6: true, nonexistent: true, couch: true*/")
+    (js2-mode)
+    (let ((found-externs (js2-get-jslint-declaration-externs)))
+      (should (and (member "Map" found-externs)
+                   (member "emit" found-externs))))))
+
+(ert-deftest js2-finds-jslint-declaration-externs-in-other-comments ()
+  (with-temp-buffer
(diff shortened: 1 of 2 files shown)


dgutov:
Hey! Thanks for the PR. Sorry I'm now just getting around to it, probably missed it back then.

Will you sign the copyright assignment to FSF? If I decide to treat the `js2-jslint-declaration-externs` declaration as data and not code, you're squeaking by under the "trivial contribution" limit, but it would be nicer to just accept it as a normal patch.

dgutov:
I'd also like to see some tests, along the lines of `test/externs.el`. With them, the patch will definitely exceed the limit. 🙁

JulianKniephoff:
No problem, I'm glad you're considering my PR for merging. :)

I will add some tests as soon as I get around to it. (Might even be today, but I won't promise anything here.)

Can you give me some pointer as to how/where to do the FSF copyright assignment? I'd be happy to do it, but I don't know how. x)

dgutov:
Great! Here are the instructions: http://orgmode.org/cgit.cgi/org-mode.git/plain/request-assign-future.txt

"The program" is Emacs, no need to change that.

JulianKniephoff:
Alright, I pushed some tests so you can review them while I wait for my answer regarding the copyright assignment, which I also just "applied for."

tests/externs.el
```diff
@@ -59,3 +59,47 @@
 ;;;TODO
 ;; ensure that any symbols bound with the import syntax are added to the extern list
 ;; ensure that any symbols bound with the export syntax exist in the file scope
+
+;; Test `/*jslint*/` declarations
+
+(ert-deftest js2-finds-jslint-declaration-externs ()
+  (with-temp-buffer
+    (insert "/*jslint browser, node:false, es6: true */")
```
dgutov: What does `node:false` mean?

JulianKniephoff: I thought I read somewhere that it had a similar meaning to something like `/*global foo: false*/`, where the assumed externs would all be read-only, but I can't find that anywhere, so maybe I dreamed it. ;)

I removed those bits from the tests and did some other improvements.

dgutov: OK. But `node: true` works? And is the same as just `node`?

The documentation examples look like

```js
/*jslint
    es6, maxerr: 10, node
*/
```

tests/externs.el
```diff
@@ -59,3 +59,69 @@
...
+    (js2-mode)
+    (should (null (js2-get-jslint-declaration-externs)))))
+
+(ert-deftest js2-finding-jslint-declaration-externs-ignores-nonexistent ()
+  (with-temp-buffer
+    (insert "/*jslint es6: true, nonexistent: true, couch: true*/")
+    (js2-mode)
+    (should (equal (delete-dups (sort (js2-get-jslint-declaration-externs) 'string<))
+                   (delete-dups (sort (append (cdr (assoc "es6" js2-jslint-declaration-externs))
+                                              (cdr (assoc "couch" js2-jslint-declaration-externs))
```
dgutov: I think this testing approach is suboptimal: if you somehow forget to include `"couch"` in `js2-jslint-declaration-externs`, this test will pass but won't check what's necessary.

Same if you make a typo here (e.g. `(assoc "coach" js2-jslint-declaration-externs)`).

tests/externs.el
JulianKniephoff: Yes, `node: true` should work and I have definitely seen this in the code that initially made me want this feature in `js2-mode`. :D

tests/externs.el
JulianKniephoff: Hm, true. Do you want me to revert to the previous method using `member`? I scrapped that because that didn't really cover everything we want to check, either. Should I just repeat all the `js2-jslint-declaration-externs` entries? :thinking: 

Or do you have a better idea? :slightly_smiling_face:

dgutov: > Do you want me to revert to the previous method using member?

Yeah, I think it was fine. Checking for one member from each of the lists seems reliable enough. If any of the lists is missing elements, it would be hard to catch with a test anyway (or the test would become really noisy and thus unreadable).

JulianKniephoff: Done and the paperwork is finished, too. :slightly_smiling_face:

dgutov: Terrific. Thanks!

PR #360, opened 2016-08-08
imenu for callback; test cases added

redguardtoo:
as title.

screen shot 2016-08-08 at 8 55 48 pm

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -60,11 +60,13 @@
 
 ;;   (add-to-list 'interpreter-mode-alist '("node" . js2-mode))
 
-;; Support for JSX is available via the derived mode `js2-jsx-mode'.  If you
-;; also want JSX support, use that mode instead:
+;; Use Emacs 27 and want to write JSX?  Then use `js2-minor-mode' as described
+;; above.  Use Emacs 26 or earlier?  Then use `js2-jsx-mode':
 
 ;;   (add-to-list 'auto-mode-alist '("\\.jsx?\\'" . js2-jsx-mode))
 ;;   (add-to-list 'interpreter-mode-alist '("node" . js2-jsx-mode))
+
+;; Note that linting of JSX code may fail in both modes.
 
 ;; To customize how it works:
 ;;   M-x customize-group RET js2-mode RET
@@ -11762,11 +11764,12 @@ Selecting an error will jump it to the corresponding source-buffer error.
     ;; Schedule parsing for after when the mode hooks run.
     (js2-mode-reset-timer)))
 
-;; We may eventually want js2-jsx-mode to derive from js-jsx-mode, but that'd be
-;; a bit more complicated and it doesn't net us much yet.
 ;;;###autoload
 (define-derived-mode js2-jsx-mode js2-mode "JSX-IDE"
-  "Major mode for editing JSX code.
+  "Major mode for editing JSX code in Emacs 26 and earlier.
+
+To edit JSX code in Emacs 27, use `js-mode' as your major mode
+with `js2-minor-mode' enabled.
 
 To customize the indentation for this mode, set the SGML offset
 variables (`sgml-basic-offset' et al) locally, like so:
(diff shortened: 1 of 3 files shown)


redguardtoo:
@dgutov , I updated the pull request. Unit test for imenu is added.

js2-mode.el
```diff
@@ -7430,21 +7441,60 @@ When passed arguments of wrong type, does nothing."
...
+     ((js2-call-node-p node)        ; foo.bar
+      (js2-compute-nested-prop-get (js2-call-node-target node)
+                                   (js2-call-node-args node)))
      ;; foo.bar.baz is parenthesized as (foo.bar).baz => right operand is a leaf
      ((js2-prop-get-node-p node)        ; foo.bar
       (setq left (js2-prop-get-node-left node)
             right (js2-prop-get-node-right node))
-      (if (setq head (js2-compute-nested-prop-get left))
-          (nconc head (list right))))
+      ;; For JS code like `foo.state(...)`':
```
dgutov: `foo.state(...)` is a function call, not a function definition. Why would we show it in the index?

redguardtoo: The is pull request is designed for anonymous callback. `foo.state(...)` is displayed as the prefix of imenu item of callback.

`foo.state(function() { console.log('hell world') });`

A common use case is `promise`,
```javascript
var p1 = new Promise();
promise.then(function(resp) {
   console.log('resp=', resp);
}).catch(function(reason) {
   console.log('reason=',reason);
});
```

With this patch, the imenu items for the above callbacks will be different

dgutov: > Looks the code in js2-mode.el can NOT be moved into js2-imenu-extras.el, or else js2-mode.el will be dependent on js2-imenu-extras.el.

If the only changes are in js2-imenu-extras.el, that won't make js2-mode.el dependent on anything new.

> My code improves walk through algorithm of the AST which some js2-mode commands are dependent on. For example, js2-jump-to-definition uses js2-compute-nested-prop-get.

What cases does its behavior improve in?

dgutov:
> js2-compute-nested-prop-get is improved to get more information from property chain. As I said, my original intention is to make imenu support for anonymous callback only

But how are you improving its original callers by that? If you don't, maybe you should create a different new function instead. Maybe that one would wrap `js2-compute-nested-prop-get`, or something.

> But by improving js2-compute-nested-prop-get, other missing imenu items could be found. Here is the screenshots.

The screenshots look unconvincing. This example:

```js
fnFactory().a = function() {  }
```

says to me that you are assigning a property on an unreachable value (because the result of the `fnFactory()` call is not assigned to anything). It looks pointless. Do you have a more realistic example?

> I think you don't need worry about the complexity of js2-mode

I do worry. It is above my preferred threshold, and it's not easy to make sense of after being away for a while. There are such places in imenu generation code.

> In my opinion, since js2-mode is a little bigger, as a lead developer, you can do some manager things. For example, you can setup technical tasks writing unit tests. I know many young talented Chinese developers would like to contribute code. I can help need resource from China Emacs community.

Thanks, but I don't have a lot of time or motivation for that. Hopefully the young developers, if they are talented enough, will get into the codebase themselves and provide helpful patches without needing too much assistance.

redguardtoo:
Thanks for asking. I will give a more practical example to answer your point 1 and point 2.

Below code is not unusual:
``` javascript
var a = {
  getb: function() {
    return this.b;
  },
  b: {
    c: 3
  }
};

a.getb(a).c = function() {
};
```
When dealing some complicated js object, it's common to use method like getb() to access the internal property.  The old `js2-compute-nested-prop-get ` algorithm will not search the complete property chain of a.getb(a).c because the function node getb() stop it. 

But as you can see, getb() is just a shortcut of property access. So there is no reason to stop the search in `js2-compute-nested-prop-get`. The new `js2-compute-nested-prop-get` will give more information.

js2-reocrd-imenu-functions, js2-jump-to-definitions, js2-search-scope use js2-compute-nested-prop-get

Here is screenshot after applying new algorithm,
new-js2-nq8

As a engineer, I told you don't worry about imenu complexity because I already did investigation.

imenu's existing algorithm can be improved to get **less code and robust result.**. I think the better algorithm is **list types of the leaf node and search from the bottom to up**. Current algorithm starts from parent of leaf node.

So it's more a algorithm problem instead of code complexity problem. I will help you to design a better algorithm before Christmas if you accept my pull request.

dgutov:
> But as you can see, getb() is just a shortcut of property access. So there is no reason to stop the search in js2-compute-nested-prop-get.

I really don't understand why you'd want to do that. And especially why you would modify the returned constant-ish value thereafter.

> js2-reocrd-imenu-functions, js2-jump-to-definitions, js2-search-scope use js2-compute-nested-prop-get

How exactly are they going to benefit from this?

> Here is screenshot after applying new algorithm,

The result doesn't look like the usual imenu structure (which more often describes the static structure of the code, while accounting for the common patterns).

But if this is your kind of thing, I think you can parse out the list of assignments simply afterwards, with regexps, similarly to how the other code in `js2-imenu.el` operates.

Issue #359, opened 2016-08-07
es7 class properties


dagda1:
I am getting syntax highlighting errors for code that uses es7 class properties:

```
import React, { Component, PropTypes } from 'react';

export default class RContainer extends Component {
  static propTypes = {
    height: PropTypes.nubmber,
    width: PropTypes.number
  };
};
```

It looks like this in emacs

![es](https://cloud.githubusercontent.com/assets/118328/17461399/0eaff96c-5c83-11e6-9769-c0cbbcf72081.png)

I can see the following message in `*messages*`

> Binding initializer not in destructuring assignment

I can see this [error](https://github.com/mooz/js2-mode/blob/master/js2-mode.el#L1742) in the js2-mode source.

Can this be recognised as a valid construct?

The code passes eslint

XeCycle:
okay that line blames to me :)
back [then](https://github.com/mooz/js2-mode/pull/280) we decided not to add ES7 features.  if we want the reverse now, I'm happy to help.

dgutov:
>  if we want the reverse now, I'm happy to help

Apparently it's stage 2 now. I wouldn't mind adding support for it.

CrackedP0t:
Babel stage 2 has it and it's a very useful feature, so I think it should be in js2-mode.

fxfactorial:
Would this also fix the syntax errors highlighting on plain public fields? ie: 

```javascript
class Foo {
   login = e => {

  }
// Also keep in mind this ability:

  bar = async e => {

  }
};
```

formido:
I get the warning on the following:

``` javascript
class App extends React.Component {
  state = {
    loading: true,
    user: null,

    // Model
    employees: [],
  }
```
> Binding initializer not in destructuring assignment

Is that still the expected behavior? I just upgraded to the latest package on MELPA.

dgutov:
All of the examples here seem to have been fixed by now.

PR #363, opened 2016-08-17
Add support for ES7 class public fields

XeCycle:
For #359.

Looking at the proposal text I am not sure whether I should allow

``` js
class C {
  x
  y
}
```

Babel allows it, but babel disagrees with the proposal in rejecting computed names, which seems fine to me.  Which one is correct?

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -10895,7 +10895,7 @@ expression)."
        ;; Found a key/value property (of any sort)
        ((member tt (list js2-NAME js2-STRING js2-NUMBER js2-LB))
         (setq after-comma nil
-              elem (js2-parse-named-prop tt previous-token))
+              elem (js2-parse-named-prop tt previous-token class-p))
         (if (and (null elem)
                  (not js2-recover-from-parse-errors))
             (setq continue nil)))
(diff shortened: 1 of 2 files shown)


dgutov:
1.1 in https://tc39.github.io/proposal-class-public-fields/ says that the initializer is optional.

Is that what you were asking?

XeCycle:
Yes I am aware, I was not sure whether auto semicolon insertion takes place here.  Of course something like `class C { x y }` is invalid, but does ASI happen here if x is followed by a newline?

dgutov:
I think so, yes. http://www.ecma-international.org/ecma-262/5.1/#sec-7.9.1 seems to imply that.

peterhadlaw:
Any status updates on this? I heavily use this feature and was wondering 👍  thanks!

XeCycle:
Looks like done to me.  This implementation allows computed property names, which are rejected by babel; again, as I said in the PR, the proposal, as I understand it, allows computed property names.  Please correct me if I was wrong.

dgutov:
OK, this looks good as well. Thank you!

Issue #362, opened 2016-08-15
Invalid "yield expression must be parenthesized." warning


bellbind:
In ECMA262, `yield` expression is one of assignment expressions.
- http://www.ecma-international.org/ecma-262/6.0/#sec-assignment-operators

Parenthesizing `yield` is required only at primary expression positions.

For example:

``` js
// wrapped paren is not required: but warned "yield expression must be parenthesized."
funcCall(yield v);
return yield v;

// wrapped paren is required ("syntax error" displayed)
a = 10 + (yield v); 
typeof (yield v);

// wrapped paren is not required (no warning)
a += yield v;
```

XeCycle:
Relative lines are said to come from very early days, when yield expression was probably some implementation extension defined differently.  IMO it should be safe to simply drop this check as of now.

dgutov:
@XeCycle What's a relative line?

You're probably right about dropping the check in `js2-parse-argument-list`, but that only seems to help with the first example. The second one still gives "syntax error".

dgutov:
Seems fixed by that PR from 2017.

PR #366, opened 2016-08-27
Fix some yield parsing (#362)

XeCycle:
`msg.yield.parenthesized` error is simply dropped, together with another fix, it parses the code example in that issue fine.  Wrapped in a generator function, of course.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -1359,9 +1359,6 @@ the correct number of ARGS must be provided."
 (js2-msg "msg.bad.yield"
          "yield must be in a function.")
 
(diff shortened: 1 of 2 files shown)


XeCycle:
sorry some other checks failed, but well, that error is no longer valid as of today.

XeCycle:
passing now.

dgutov:
Thanks! Could you add a couple of tests?

XeCycle:
Added a fix for #388.

dgutov:
Thanks!

PR #377, opened 2016-10-09
Special methods should allow keywords as names

XeCycle:
Noticed `{ async delete() {} }` parsed incorrectly, and this is a fix.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -7734,14 +7734,14 @@ is only true until the node is added to its parent; i.e., while parsing."
         ;; Tell cc-engine the bounds of the comment.
         (js2-record-text-property beg (1- end) 'c-in-sws t)))))
 
-(defun js2-peek-token ()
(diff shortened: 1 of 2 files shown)


js2-mode.el
```diff
@@ -7821,6 +7821,9 @@ string is NAME.  Returns nil and keeps current token otherwise."
 (defun js2-get-prop-name-token ()
   (js2-get-token (and (>= js2-language-version 170) 'KEYWORD_IS_NAME)))
 
+(defun js2-peek-prop-name-token ()
+  (js2-peek-token (and (>= js2-language-version 170) 'KEYWORD_IS_NAME)))
```
dgutov: OK, I see why you did that, but I'd probably inline this function at its only call site.

XeCycle:
done.

dgutov:
Thanks!

PR #378, opened 2016-10-09
Compute externs on demand

### LINKED ISSUE #375: Setting `js2-include-*-externs` in `.dir-locals.el` has no effect
When you work on different JavaScript projects with mixed environments (say one for the browser, another one with a Node.js environment), you might not want to globally set the `js2-include-*-externs` variables in your Emacs configuration.

One solution normally would be to set them per project in a `.dir-locals.el` file, for example like this:

```elisp
((js2-mode . ((js2-include-node-externs . t))))
```

This does not work however, since these variables only affect `js2-mode`s initialization, apparently, which happens before evaluating the relevant section of the `.dir-locals.el`.

When I first ran into this issue, googling around for a solution brought up [this](https://www.reddit.com/r/emacs/comments/4mzz6q/js2mode_and_dirlocalsel/) thread on reddit, which has some workarounds and also explains the issue nicely. But no one seems to have reported it here, so I went ahead and did it.

If this is not easily fixable, a hint on how to do it right in the documentation for the relevant variables might be a good interim solution, maybe.

JulianKniephoff:
Instead of only taking into account the `js2-include-?-externs`
variables when initializing the mode, compute what was
`js2-default-externs` before every time we highlight undeclared
identifiers.

Closes #375

### Some notes

In the documentation for `js2-default-externs` it says that **at the moment** the externs are only used to highlight undeclared variables, so I thought a bit about how I can keep my changes very general so that one might use the resulting code for other purposes as well. Since I don't know the code very well, though, I decided YAGNI (for now) and went for the simplest solution (the first one in #375).

I also refactored the documentation a bit correspondingly.

Note that a consequence of these changes is that, other than by overriding `js2-ecma-262-externs` itself, you can now no longer exclude that list from the list of externs.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -115,7 +115,7 @@
           decodeURI decodeURIComponent encodeURI
           encodeURIComponent escape eval isFinite isNaN
           parseFloat parseInt undefined unescape))
-"Ecma-262 externs.  Included in `js2-externs' by default.")
+"Ecma-262 externs.  Never highlighted by `js2-highlight-undeclared-vars'.")
 
 (defvar js2-browser-externs
   (mapcar 'symbol-name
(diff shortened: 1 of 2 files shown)


dgutov:
Thanks!

> In the documentation for js2-default-externs it says that at the moment the externs are only used to highlight undeclared variables, so I thought a bit about how I can keep my changes very general so that one might use the resulting code for other purposes as well.

It's an old text. I don't think there's much else we could use this exact data for, so there's no need to worry about it.

> Note that a consequence of these changes is that, other than by overriding js2-ecma-262-externs itself, you can now no longer exclude that list from the list of externs.

That's probably okay. If someone complains, we can split this logic to occur before `js2-post-parse-callbacks` are run.

Could you also add a NEWS entry?

js2-mode.el
```diff
@@ -7092,14 +7074,33 @@ later. NODE must be a name node."
...
+- the list `js2-ecma262-externs' for basic names from the ECMAScript language standard
+- depending on the buffer-local variables `js2-include-?-externs'
+  the corresponding `js2-?-externs' to add names for certain environments
+  like the browser or node/rhino
+- two user-customizable lists `js2-global-externs' and `js2-additional-externs'
+  of which the latter is a buffer-local variable
+
+See especially `js2-additional-externs' for further details about externs."
+  (let ((default-externs
+          (append js2-ecma-262-externs
```
dgutov: `js2-ecma-262-externs` or `js2-browser-externs` should probably go last here (the last argument is not copied).

This function is probably fast enough anyway, but a tiny speed boost wouldn't hurt.

dgutov:
Thanks!

PR #381, opened 2016-10-15
Fix "pos" argument in template node constructors

mgiles:
Currently the make-js2-template-node and make-js2-tagged-template-node constructors take a "beg" argument, which does not set the js2-node "pos" field. The pos field not being set causes incorrect results from, for example, js2-node-at-point with the point inside a template literal placeholder.

The only change here is to rename "beg" to "pos" in the constructors and call sites. 

I'm happy to make any changes needed to get this merged. Please let me know if this change needs any new test cases.


dgutov:
Thanks! We don't usually test the value of pos, so let's do without new tests, for now.

PR #384, opened 2016-10-25
Fix length of name in namespace-import nodes

mishoo:
Should be `(- (js2-current-token-end) (js2-current-token-beg))` rather
than just `(js2-current-token-end)`.

Also change `if` to `cond` (seems easier to read in this case).


dgutov:
Thanks!

mishoo:
Thanks for merging.  Here's another one: https://github.com/mooz/js2-mode/pull/385 (sorry for the regression)

Issue #383, opened 2016-10-25
determine whether previous node is comment node


futurist:
When have a normal node, e.g. an array element node, want to find whether there's some comment node before this element, I've tried but none of them worked, as below:
1. **js2-comment-node-p (pos)**

It accept **pos**, but only have the array element's pos.
1. **js2-node-find-child-before (pos, parent)**

Want to find the array element's previous node, but

> ﻿PARENT is any node supported by ‘js2-node-child-list’.

so the search failed for a `comment node`

Any solution for this in js2-mode? Thanks!

dgutov:
Is the input value to the desired function a node, or a buffer position?

futurist:
```lisp
(defun test-comment()
  (let ((node (js2-node-parent (js2-node-at-point)))
        (pos (point)))
    (js2-node-short-name (js2-node-find-child-before pos node))))
```

```js
var a = [
  1,
  //sdf
  2|,
  3
]
```

put your point onto **2**, and eval `(test-comment)`, the result always `js2-number-node`

futurist:
**Another problem**

When tested with below code, emacs will die every time:

```lisp
(defun test-comment()
  (let ((node (js2-node-parent (js2-node-at-point)))
        (pos (point)))
    (js2-node-abs-pos (js2-node-find-child-before pos node))))
```

```js
/*
  Copyright @ Michael Yang
  License MIT
*/

// phantomjs module
var fs = require('fs')
var sys = require('system')
var helper = require('./helpers/helper.js')
var page = require('webpage').create()

var DEBUG = sys.env['DEBUG']
var debug = function (arg) {
  var args = [].slice.call(arguments)
  if (DEBUG) console.log.apply(console, args)
}

var a= [
  {
    a:1
    ,
    b:2
  }
  ,
  [
    2
    ,
    3
  ],
  //sdf
  4,
  5
]

var b=[
  1,
  //sdf
  2,   //<<============point position under 2
  3
]
```

dgutov:
Hi!

Looking into it, the actual problem (in the example 1) is that comments are not part of the AST. Or rather, there is a list of comments attached to the AST root (which you can access with `js2-ast-root-comments`), but the comment node is not one of the children of the array node. To find the closes comment node, you can take that list and do a linear search on it. See `js2-comment-at-point` as one example.

Next, you could use the proposed `test-comment` function and find the closest non-comment node, to check whether the comment node is closer.

dgutov:
[...]

PR #387, opened 2016-11-03
feat: add js2-comments-between func

futurist:
See issue https://github.com/mooz/js2-mode/issues/383

Will squash into 1 commit if the PR become mergable.


tests/consume.el
```diff
@@ -0,0 +1,59 @@
+;;; tests/consume.el --- Some tests for js2-mode.
+
+;; Copyright (C) 2009, 2011-2016  Free Software Foundation, Inc.
```
dgutov: Could you fix the dates? These should only list the years the file has seen any changes. Just 2016 in this case, IOW.

dgutov:
Thanks!

Issue #260, opened 2015-09-02
Allow ignoring specific warnings.


jordonbiondo:
From what I see there isn't a way to ignore specific warnings from js2-mode. There are certain times (usually in a test file) where I would like to ignore false warnings brought up by js2-mode.

For instance in a test file using chai's expect. I may have a line like this:

```
expect(true && false).to.be.false;
```

These lines are always highlighted because js2-mode thinks the line has no side effects even though they actually do.

My suggestion would be to have a buffer-local variable that contains a list of warnings to ignore. And then in `js2-report-warning` check if the warning is in the list before proceeding.

That way I can do something like this in my config in order to solve my problem.

```
(add-hook 'js2-mode-hook
                 (lambda ()
                   (when (is-a-test-file)
                     (add-to-list 'js2-var-with-a-list-of-ignored-warnings "msg.no.side.effects"))))
```

If this could be added to `js2-mode` I'd be happy to submit a pull request. But given the size of the project, there may be a better way to do this I overlooked.

jordonbiondo:
I have this implemented in my config like this:

```lisp
    (defvar-local jordon-js2-ignored-warnings nil)

    (defadvice js2-report-warning (around ignore-some-warnings activate)
      (unless (member (ad-get-arg 0) jordon-js2-ignored-warnings)
        ad-do-it))
```
Then in a hook I add certain warnings to the list depending on the type of file.

dgutov:
Another option is to add a function to `js2-post-parse-callbacks` which would filter out undesired elements from `js2-parsed-errors` and `js2-parsed-warnings`.

decoursin:
bump

RockyRoad29:
Some details about possible solutions would be welcome on emacswiki or elsewhere with a pointer here ... 
I am also trying to get rid of those annoying warnings on statements like:
http://chaijs.com/api/bdd/#method_true
Thanks

Issue #355, opened 2016-07-23
disable warning `msg.no.side.effects`?


blaenk:
I'm using chai for testing my code but certain constructs make js2-mode think that there are no side-effects. I think the side-effects warning is useful otherwise, but it's noisy in this context. Is there a way to disable it when it recognizes it's in an `expect(_)` chain? Failing that, is there a way to _only_ disable this warning message but not all other warning messages?

![screenshot](https://i.imgur.com/N6wSf3M.png)

jordonbiondo:
For those looking for this feature right now, refer to issue #260 where I have code that does exactly this.

Also take a look at my config for js2-mode where my complete warning ignore code exists: https://github.com/jordonbiondo/.emacs.d/blob/master/init.el#L901

It would be great if this was built in to the mode. I'd be happy to open a PR that makes this a real feature if the maintainers are interested in the idea.

aspiers:
See https://github.com/mooz/js2-mode/pull/404#issuecomment-997100520 for a quick and easy way to apply this to a whole `test/` subdirectory.

PR #404, opened 2017-01-27
Allow specific warnings to be ignored

jordonbiondo:
The need for disabling specific warning messages has come up a few times.

I've seen that there have been changes made to support disabling specific warnings such as dangling commas and no semicolons, but there's not a blanket method for disabling anything you want.

This was discussed shortly in #260, but a conclusion wasn't reached.

This is a much smaller first-class version of the functionality I've built in my own config: https://github.com/jordonbiondo/.emacs.d/blob/master/init.el#L901

Let me know what you think about the possibility of this feature being accepted using this method or any other.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -226,6 +226,13 @@ variable with predicate PRED."
   (make-variable-buffer-local name)
   (put name 'safe-local-variable pred))
 
+(defcustom js2-ignored-warnings nil
+  "A list of warning message types that will not be reported.
+
+Possible values are the keys of `js2-message-table'."
+  :group 'js2-mode
+  :type '(repeat string))
+
 (defcustom js2-highlight-level 2
   "Amount of syntax highlighting to perform.
 0 or a negative value means none.
@@ -8130,10 +8137,22 @@ Scanner should be initialized."
     (setq js2-mode-ast root)  ; Make sure this is available for callbacks.
     ;; Give extensions a chance to muck with things before highlighting starts.
     (let ((js2-additional-externs js2-additional-externs))
+      (js2-filter-parsed-warnings)
       (save-excursion
         (run-hooks 'js2-post-parse-callbacks))
       (js2-highlight-undeclared-vars))
     root))
+
+(defun js2-filter-parsed-warnings ()
+  "Remove `js2-parsed-warnings' elements that match `js2-ignored-warnings'."
+  (when js2-ignored-warnings
+    (setq js2-parsed-warnings
+          (cl-remove-if
+           (lambda (warning)
+             (let ((msg (caar warning)))
+               (member msg js2-ignored-warnings)))
+           js2-parsed-warnings)))
+  js2-parsed-warnings)
 
 (defun js2-parse-function-closure-body (fn-node)
   "Parse a JavaScript 1.8 function closure body."


jordonbiondo:
With this feature, I'd scrap much of what is in my current js2-mode config, and replace it with this:

```lisp
(with-eval-after-load "js2-mode"

  (setq-default 
   js2-ignored-warnings 
   '("msg.return.inconsistent" "msg.anon.no.return.value" "msg.no.return.value"))

  (make-variable-buffer-local 'js2-ignored-warnings)

  (defun my-js2-test-file-setup ()
    (when (string-match-p "_test.js$" (buffer-file-name))
      (push "msg.no.side.effects" js2-ignored-warnings)))

  (add-hook 'js2-mode-hook 'my-js2-test-file-setup))
```

This code would also satisfy the needs of @blaenk in #355

js2-mode.el
```diff
@@ -226,6 +226,13 @@ variable with predicate PRED."
   (make-variable-buffer-local name)
   (put name 'safe-local-variable pred))
 
+(defcustom js2-ignored-warnings nil
+  "A list of warning message types that will not be reported.
+
+Possible values are the keys of `js2-message-table'."
+  :group 'js2-mode
+  :type 'list)
```
dgutov:
```lisp
:type '(repeat string))
```

maybe?

js2-mode.el
```diff
@@ -8130,11 +8137,23 @@ Scanner should be initialized."
...
     root))
 
+(defun js2-filter-parsed-errors ()
+  "Remove `js2-ignored-warnings' from `js2-parsed-errors'."
+  (when js2-ignored-warnings
+    (setq js2-parsed-warnings
+          (cl-remove-if
+           (lambda (warning)
+             (let ((msg (caar warning)))
+               (and msg (member msg js2-ignored-warnings))))
```
dgutov: I think `(and msg` is unnecessary here. And are there any warnings with empty message anyway?

js2-mode.el
```diff
@@ -8130,11 +8137,23 @@ Scanner should be initialized."
     (setq js2-mode-ast root)  ; Make sure this is available for callbacks.
     ;; Give extensions a chance to muck with things before highlighting starts.
     (let ((js2-additional-externs js2-additional-externs))
+      (js2-filter-parsed-errors)
       (save-excursion
         (run-hooks 'js2-post-parse-callbacks))
       (js2-highlight-undeclared-vars))
     root))
 
+(defun js2-filter-parsed-errors ()
```
dgutov: Shouldn't it be called `js2-filter-parsed-warnings`?

js2-mode.el
```diff
@@ -8130,11 +8137,23 @@ Scanner should be initialized."
...
+  "Remove `js2-ignored-warnings' from `js2-parsed-errors'."
```
dgutov: Better:

Remove `js2-parsed-warnings` elements that match `js2-ignored-warnings`.

blaenk:
Just noticed the ping on #355. Thanks for your work on this, I'm sure many will find this useful, including myself.

js2-mode.el
jordonbiondo: I'm not familiar enough with `defcustom` types to comment on the "maybe", I'lll have to defer to you.

jordonbiondo:
@dgutov I've updated the commit with all your suggestions, let me know if there's anything else I can do.

js2-mode.el
dgutov: One way to review this kind of suggestion is to make the change and see if the `customize-variable` UI looks no worse (or probably better) than before.

Anyway, it must be fine.

dgutov:
@jordonbiondo Thanks. Have you contributed to Emacs before?

This patch meets the allowed limit, but any further patches will require signed copyright assignment papers.

jordonbiondo:
@dgutov No I have not signed the assignment papers.

jordonbiondo:
I am in the process up signing the assignment paperwork now.

dgutov:
Thanks!

aspiers:
BTW just in case anyone is reading this years later, this feature can also be utilised by placing something like the following in `.dir-locals.el`:

```elisp
 ("test"
  . ((js2-mode
      . ((js2-ignored-warnings . ("msg.no.side.effects"))))))
```

Then it would apply to the `test/` subdirectory.

PR #410, opened 2017-03-29
Allow trailing comma in param list

### LINKED ISSUE #403: Dangling or trailing commas in parameter list
ES6 proposal allows a dangling comma in function parameter list. Is there a way to get js2 to be less strict about warnings on this as currently it parses as a syntax error:

```js

function foo(a,
             b,
             c,
             )
```

ttmarek:
Fixes #403 

I'd like to add a unit test for this, but I don't really know where to start since I'm new to emacs lisp and emacs mode development. There doesn't seem to be a unit test for `js2-parse-function-params` in `tests/parse.el`. Do you think its worth adding one?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # History of user-visible changes
 
+## Next
+
+* Support for trailing commas in function parameter lists.
+
 ## 2017-01-16
 
 * `js2-include-*-externs` are now evaluated on demand.  As a result,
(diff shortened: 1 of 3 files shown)


dgutov:
> I'd like to add a unit test for this, but I don't really know where to start

A high-level test would be much easier to write. See all the `js2-deftest-parse` declarations in `parse.el`, e.g. `octal-number-okay`.

ttmarek:
Added a test. Its pretty simple. Welcome for feedback if it could be any better. Basically we're expecting that `function foo(a, b,) {\n}` won't return a syntax error.

dgutov:
Thanks!

Could you add a NEWS entry as well? At the top, under heading "Next".

PR #421, opened 2017-04-28
Map C-M-j to js2-line-break

raxod502:
In Emacs, `indent-new-comment-line` is bound to both `M-j` and `C-M-j`. `js2-mode` remapped `M-j` to `js2-line-break`, but didn't remap `C-M-j`. This commit fixes that by telling Emacs to remap *all* bindings for `indent-new-comment-line` (currently, just `M-j` and `C-M-j`) to `js2-line-break`.


dgutov:
Okay, thanks!

PR #424, opened 2017-05-15
Add option js2-getprop-has-side-effects

shicks:
Includes a slight restructuring of how side-effect calculations are done (for better testability) and adds basic test coverage for no.side.effect warnings.

### DIFF
--- a/tests/parser.el
+++ b/tests/parser.el
@@ -1251,3 +1251,40 @@ the test."
 (js2-deftest-classify-variables destructure-object-mixed
   "function foo() { let {a, b, c = 3} = {a: 1, b: 2}; }"
   '("foo@10:U" "a@23:U" "b@26:U" "c@29:U"))
+
+;; Side effects
+
+(js2-deftest no-side-effects-at-top-level
+  "var x; x.foo;"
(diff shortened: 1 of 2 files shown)


tests/parser.el
```diff
@@ -1251,3 +1251,40 @@ the test."
 (js2-deftest-classify-variables destructure-object-mixed
   "function foo() { let {a, b, c = 3} = {a: 1, b: 2}; }"
   '("foo@10:U" "a@23:U" "b@26:U" "c@29:U"))
+
+;; Side effects
+
+(js2-deftest no-side-effects-at-top-level
+  "var x; x.foo;"
+  (js2-mode--and-parse)
+  (should (equal nil js2-parsed-warnings)))
```
dgutov: This can be more succinctly written as `(null js2-parsed-warnings)`.

dgutov:
Looking good, thanks. Just please make the small alteration in the tests (there are existing examples using `(equal nil ...)`, but we'll get to them later).

tests/parser.el
shicks: Done.

dgutov:
Thanks!

PR #425, opened 2017-05-16
Fix jsdoc highlighting for tags w/ optional braces

shicks:
Some of the tags in js2-jsdoc-arg-tag-regexp (specifically `@this`, `@throws`, and `@type`) are used with braces by Closure Compiler. Previously the close-brace was styled differently from the open-brace.  Adding a branch to the regex for optional braces fixes this.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -6943,7 +6943,7 @@ of a simple name.  Called before EXPR has a parent node."
              "throws"
              "type"
              "version"))
-          "\\)\\)\\s-+\\([^ \t]+\\)")
+          "\\)\\)\\s-+\\([^ \t\n]+\\)")
   "Matches jsdoc tags with a single argument.")
 
 (defconst js2-jsdoc-empty-tag-regexp


dgutov:
Could you explain the patch some more? I see the problem, but the change below also seems to fix it:

```diff
diff --git a/js2-mode.el b/js2-mode.el
index 0389833..e785ea0 100644
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -6951,7 +6951,7 @@ of a simple name.  Called before EXPR has a parent node."
              "throws"
              "type"
              "version"))
-          "\\)\\)\\s-+\\([^ \t]+\\)")
+          "\\)\\)\\s-+\\([^ \t\n]+\\)")
   "Matches jsdoc tags with a single argument.")
 
 (defconst js2-jsdoc-empty-tag-regexp
```

shicks:
I don't really understand what's going on there - how the braces are pulled out into a different font, nor why my patch fixed it.  It looks like yours works, too.  That approach seems better, TBH.  I've updated the PR.

dgutov:
Thanks.

Could you also avoid ending the summary (first line of the message) with the period?

http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

shicks:
Done.

PR #432, opened 2017-07-01
Add smart-tabs-mode to "see also" in README

strugee:


dgutov:
Sorry but no. The "see also" packages use the JS2 mode AST is some way or other.

It's not just a list of random recommendations.

strugee:
@dgutov yeah, I read the text. AFAIK smart-tabs-mode _does_ use the AST.

strugee:
Ah, I just checked and they don't use it after all. Sorry for the noise :/

PR #436, opened 2017-07-21
Always consider `await` to have side-effects

### LINKED ISSUE #429: await incorrectly described as having no side effects
Sometimes it's convenient or necessary to create promises before awaiting them, which results in code like the following:

```js
async function example() {
  const p1 = queryDB();
  const p2 = anotherDBQuery();
  await p1;
  await p2;
}
```

Currently the two awaits are flagged as having no side-effects, which is incorrect.

felipeochoa:
Closes #429


dgutov:
Thanks!

PR #440, opened 2017-08-04
make sure to visit import clause children in order

### LINKED ISSUE #439: import-clause children not visited in order
If you have an import statement with both a default import and named imports, `js2-visit-import-clause` visits children in the wrong order. Based on the parsing rules in `js2-mode` (and  [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)), the default import should always come first.

The impact of this bug is that `js2-node-at-point` does not pick up on the `js2-name-node` in these cases. For example:

```js
import defaultImport, { a, b, c} from 'xyz';
```

If you put point in `defaultImport` and `M-x eval-expression (js2-node-short-name (js2-node-at-point))`, you get `"js2-import-clause-node"`

felipeochoa:
Fixes #439

### DIFF
--- a/tests/consume.el
+++ b/tests/consume.el
@@ -66,7 +66,6 @@
     (js2-mode--and-parse)
     (let (visit-log)
      (js2-visit-ast js2-mode-ast (lambda (node end-p)
-                                   (message (js2-node-short-name node))
                                    (when (and (not end-p) (js2-name-node-p node))
                                      (let* ((start (js2-node-abs-pos node))
                                             (end (+ start (js2-node-len node))))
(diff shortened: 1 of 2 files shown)


tests/consume.el
```diff
@@ -57,3 +57,20 @@
...
+
+;;; Visitors
+
+(ert-deftest js2-visit-import-clause-in-order ()
+  (with-temp-buffer
+    (insert "import defaultImport, { a, b, c} from 'xyz';")
+    (js2-mode--and-parse)
+    (let (visit-log)
+     (js2-visit-ast js2-mode-ast (lambda (node end-p)
+                                   (message (js2-node-short-name node))
```
dgutov: Should we remove the `message` call here?

felipeochoa: Woops. Fixed

PR #442, opened 2017-08-12
Correctly handle arrow functions in js2-node-parent-stmt

felipeochoa:
Arrow functions are always expressions, so `js2-node-parent-stmt` should
continue recursing to their parent to find the first statement. This could happen
either because it is called with the arrow function node to begin with, or if the
arrow function is brace-less, in which case it doesn't wrap its expression in `js2-expr-stmt-node`


dgutov:
Thanks!

PR #464, opened 2017-11-01
Set the correct node pos for unary nodes

felipeochoa:
Currently, prefix unary nodes have their `pos` attribute set to the start of their argument rather than the start of the operator. This causes issues with `js2-node-at-point` and similar. This PR goes through all these node creation sites and sets the correct attribute. I guess it's technically a breaking change since it adds a non-optional argument to `js2-make-unary`, though that strikes me as an internal function.


dgutov:
> that strikes me as an internal function

Me too.

Thanks!

PR #466, opened 2017-11-02
Fix length of shorthand object property nodes (e.g. `{foo,bar}`)

mishoo:
Also, drop `(constructor nil)` on structures where a constructor is defined.


mishoo:
Any concerns about this one?

PR #467, opened 2017-11-03
Highlight single line jsdoc comments

felipeochoa:
E.g.,

```js
/** @implements {Interface} */
var f = // ...
```


dgutov:
So JSDoc parsers support this form, too?

felipeochoa:
Not sure about the others, but typescript's does

dgutov:
Good enough, thanks!

PR #469, opened 2017-11-06
Add @typedef jsdoc highlighting support

felipeochoa:
There's a little bit of a hack: [`@typedef`](usejsdoc.org/tags-typedef.html) is highlighted here the same way as [`@param`](http://usejsdoc.org/tags-param.html), though the grammars are technically different.

The grammar for `@typedef` is `@TAG [OPTIONAL-TYPE] NAME`, whereas `@param` technically has one extra production for the optional description (which js2 doesn't actually match on or highlight)


dgutov:
OK, I'll defer to your understanding here. :+1:

PR #473, opened 2017-11-28
Fix error when parsing malformed namespace import

felipeochoa:
Malformed import statements cause the parser to error:

```js
import * lib from 'lib';
```

This patch doesn't add anything in terms of parse-error recovery (the above has 7 parse errors), but it prevents the parser from crashing. I did run a couple of checks and the parser seems to always recover after the quoted source file.


dgutov:
Thanks!

PR #487, opened 2018-03-04
Classify imported variables

lelit:
This handles most common cases of the `import` statement, see the tests.
I would not be surprised if there is a wider set of syntaxes... if so, please give me samples and I will try to cover them too.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -7192,58 +7192,71 @@ key of a literal object."
         (let ((left (js2-prop-get-node-left parent)))
           (when (js2-name-node-p left)
             (js2--add-or-update-symbol left nil t vars)))
-      (let ((granparent parent)
-            var-init-node
-            assign-node
-            object-key         ; is name actually an object prop key?
-            declared           ; is it declared in narrowest scope?
-            assigned           ; does it get assigned or initialized?
-            (used (null function-param)))
-        ;; Determine the closest var-init-node and assign-node: this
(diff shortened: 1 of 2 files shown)


azazel75:
The complete syntax of the import statement is here http://www.ecma-international.org/ecma-262/6.0/#sec-imports

You may want to signal as error an import like:
```javascript
import {default, foo} from 'module';
```
when the special name `default` is involved in named imports, it should always be aliased, i.e.:
```javascript
import { default as bar,  foo} from `module`;
```

azazel75:
mmm sorry, it doesn't appear to be formally forbidden... maybe a warning will suffice...

lelit:
Thanks for the pointer!
The context of the proposed change is the analysis of the variables usage, the actual parse is done earlier in the process, though. If needed, that warning should be emitted at that time.

azazel75:
oh, maybe I'm completely wrong... it's mentioned on [ExploringJS](http://exploringjs.com/es6/ch_modules.html#sec_importing-exporting-details) but not on the [mozilla reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)

dgutov:
Thanks!

> I would not be surprised if there is a wider set of syntaxes... if so, please give me samples and I will try to cover them too.

TBH, I'm not much of an expert on JS syntax these days.

lelit:
Thank you Dmitry!

PR #491, opened 2018-03-29
Use regular forward-sexp when inside strings

felipeochoa:
There's a strange inconsistency when using `forward-sexp` and `backward-sexp`: Moving backwards takes you to the beginning of the string, whereas moving forwards skips over a symbol or balanced pair.

This patch makes `js2` more consistent with other modes by disabling its custom `forward-sexp` inside strings.


dgutov:
Sounds good. Thanks!

PR #492, opened 2018-04-01
Allow safe additional-externs as file/dir-local

felipeochoa:
Not 100% sure this is correct, but I'm trying to set a couple of safe externs from a .dir-locals
file.


dgutov:
LGTM

PR #496, opened 2018-05-20
Fix #495 Aggressively reparse in js2-mode-create-imenu-index if possible

### LINKED ISSUE #495: js2-mode-create-imenu-index no AST under racing condition
`js2-mode-create-imenu-index` currently only returns an index when there's an AST, but when the file has just been loaded and the first parse has not finished, there will be no imenu index, even though eventually it will produce an index on the next call once the parsing has finished. I don't believe there is a situation where it is desirable to return nil from `js2-mode-create-imenu-index` because the first call hasn't finished. If performance may be an issue, the user may be well advised to not turn on imenu by issuing `imenu-add-menubar-index`, or introduce a new js2 option initialized to `imenu-auto-rescan-maxout`, such that `js2-mode-create-imenu-index` to return nil when buffer size is > it.

wyuenho:
Aggressively reparse when AST is not available during first call to imenu-auto-rescan-maxout

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -12577,8 +12577,11 @@ destroying the region selection."
         (goto-char (cl-cadadr e))))))
 
 (defun js2-mode-create-imenu-index ()
-  "Return an alist for `imenu--index-alist'."
-  ;; This is built up in `js2-parse-record-imenu' during parsing.
+  "Returns an alist for `imenu--index-alist'. Returns nil on first
+scan if buffer size > `imenu-auto-rescan-maxout'."
+  (when (and (not js2-mode-ast)
+             (<= (buffer-size) imenu-auto-rescan-maxout))
+      (js2-reparse))
   (when js2-mode-ast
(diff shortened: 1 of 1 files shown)


js2-mode.el
```diff
@@ -12577,8 +12577,11 @@ destroying the region selection."
         (goto-char (cl-cadadr e))))))
 
 (defun js2-mode-create-imenu-index ()
-  "Return an alist for `imenu--index-alist'."
-  ;; This is built up in `js2-parse-record-imenu' during parsing.
+  "Returns an alist for `imenu--index-alist'. Returns nil on first
+scan if buffer size > `imenu-auto-rescan-maxout'."
+  (when (and (<= (buffer-size) imenu-auto-rescan-maxout)
```
dgutov: Maybe put `(not js2-mode-ast)` first?

dgutov:
Looks good otherwise.

wyuenho:
Done

PR #498, opened 2018-06-26
Fix broken js2-node-short-name for Emacs 26

wyuenho:
`cl-defstruct` no longer prefix struct names with `cl-struct-` in Emacs 26, this
has broken a couple of tests in [`rjsx-mode`](https://github.com/felipeochoa/rjsx-mode/issues/54), this PR fixes it.


wyuenho:
Hmm CI's can't install emacs-snapshot, @dgutov can you give it look?

PR #501, opened 2018-07-19
fix for nav into shorthand prop:value syntax

thorrr:
fix for broken nav:

```javascript
function start(n) {
    return n+1;
}

module.exports = {start};
// put cursor here  ^
```

Doing `js2-jump-to-definition` here results in an error:  `js2-jump-to-definition: Wrong type argument: arrayp, nil`

The fix is to check for the parent node's `SHORTCUT ` property set which indicates the object literal is using the new abbreviated property:value syntax.  The old code only checks to see if the current node is `parent.left`, which is also true if using abbreviated syntax.


dgutov:
Thank you!

Issue #490, opened 2018-03-20
js2-jsx-mode - Issue with intendation after code block in jsx


kbrah:
I'm having a problem with intendation in js2-jsx-mode. The problem occurs when there is jsx next to a code block

expected:
```
<App>
    <div>
        {variable1}
        <Component/>
    </div>
</App>
```
actual behavior:
```
<App>
    <div>
        {variable1}
        <Component/>
</div>
</App>
```
I'm using spacemacs and emacs v25.3 on Windows.
Is this a known issue or am I doing something wrong? (I'm fairly new to the emacs world)

dgutov:
Indentation problems with JSX are known and expected, and so far there are more reports than people interested in fixing them (sorry).

Also see https://github.com/mooz/js2-mode/#bugs

kirubakaran:
@kbrah I've found that wrapping the contents of the `{}` expression in `()` works most of the time. I'm trying to help fix the indentation issue though. So can you please reproduce the issue in a separate file and share that with me? Thanks!

wyuenho:
`js-mode` 's `js-jsx-indent-line` was originally implemented in this project and later pushed upstream.  I think trying to fix it without looking at `js2-mode`'s AST is just hopeless as the heuristics is super brittle. I've fixed this issue and a number of annoyances in `rjsx-mode`, please check it out there.

P.S. @dgutov have you thought about removing E4X from `js2-mode` and merging `js2-mode` and `rjsx-mode` ?

steinarb:
I'm seeing this with rjsx-mode 20180625.58 which pulls in js2-mode 20180605.1425 (from melpa) and I still see this issue every time I use a map on an array to generate a list.

Eg. this:
```
return (
    <div>
        <form onSubmit={ e => { e.preventDefault(); }}>
            <div className="container">
                <div className="row">
                    <div className="col">
                        <label htmlFor="username">Velg jobb:</label>
                    </div>
                    <div className="col-xs">
                        <select onChange={(event) => onJobtypeFieldChange(event.target.value, jobtypesMap, account, performedjob)} value={performedjob.transactionName}>
                            {jobtypes.map((val) => <option key={val.id}>{val.transactionTypeName}</option>)}
                        </select>
                    </div>
                </div>
                ... 
```
becomes this if I reindent
```
return (
    <div>
        <form onSubmit={ e => { e.preventDefault(); }}>
            <div className="container">
                <div className="row">
                    <div className="col">
[...]

Issue #482, opened 2018-02-15
js2-jsx-mode - controlling indent for the closing part of a self-closing tag


nukeop:
Right now the indentation for self closing tags that take multiple lines looks something like this:
```javascript
      <Tag

	 />
```

I want it to be aligned to the opening part of the tag, like this:

```javascript
      <Tag

      />
```

How do I achieve that? Setting js2-basic-offset or sgml-basic-offset or sgml-attribute-offset doesn't help.

Issue #462, opened 2017-10-24
js2-jsx-mode indent bug when nesting jsx in render={} attribute


Juul:
Everything after the first `<Route />` with jsx embedded in a `render={}` attribute is indented wrong:

```
      return (
        <Router>
          <Bar>
            <Route exact path="/foo" render={() => (
              <div>nothing</div>
            )} />
          <Route exact path="/bar" />
        </Bar>
          </Router>
      )
```

Issue #459, opened 2017-10-10
Issue with `js2-jsx-mode` indentation for self closing components with JS blocks for props


domtronn:
Hi there, I have some JSX in the following form

```jsx
export default ({ stars }) => (
  <div className='overlay__container'>
    <div className='overlay__header overlay--text'>
      Congratulations!
    </div>
    <div className='overlay__reward'>
      <Icon {...createIconProps(stars > 0)} size='large' />
        <div className='overlay__reward__bottom'>
          <Icon {...createIconProps(stars > 1)} size='small' />
            <Icon {...createIconProps(stars > 2)} size='small' />
        </div>
    </div>
    <div className='overlay__description overlay--text'>
      You've created <large>1</large> reminder
    </div>
  </div>
)
```

And the indentation is off for the `Icon` components when calling `js2-jsx-indent-line`... Doing a bit of experimentation, I found that removing the JavaScript block _or_ closing the component tag explicity, _i.e.

```jsx
<Icon size='large' />
<Icon {...createIconProps(stars > 1)} size='large'></Icon>
```

Works correctly for the indentation. However, I don't want to explicitly close the components when I don't have to... I guess during parsing it's recognising the self closing tag as an open XML tag for some reason when it contains `{}`?

jozefpetro:
Same problem here but it's happening only when using class-based component
<img width="592" alt="screen shot 2017-12-19 at 16 14 46" src="https://user-images.githubusercontent.com/15748491/34163709-c3bc3fea-e4d7-11e7-8db1-870a98ca6d42.png">

dgutov:
@jozefpetro I can't reproduce your problem. It likely depends on the surrounding code that's not in your screenshot.

Either way, `js2-jsx-mode` uses `js-jsx-indent-line` for indentation, so this report should go to the Emacs core.

PR #523, opened 2019-04-09
Use JSX highlighting and indentation added in Emacs 27

### LINKED ISSUE #140: Highlighting & indenting embedded XML (E4X)?

### LINKED ISSUE #409: JSX quotation bug
When I use `'` in the JSX tags like

``` JSX
function Bug() {
    return <div>C'est Montréal</div>;
}
```

syntax highlighting and indenting, .... are broken.

Is is possible solving this issue?

jacksonrayhamilton:
The `master` branch of Emacs now provides support for highlighting JSX, greatly improves the current JSX indentation support, and automatically detects the use of JSX (without needing to enter `js-jsx-mode`).  These changes allow js2-mode to tap into these improvements when using Emacs on the `master` branch, or eventually in version 27.

Fixes #140, #330, #389, #411, #451, #459, #462, #482, #490

~~I was hoping this would address #409, but I think we may have a little more work to do for that.  Indentation is right in js-mode, but I think js2-mode is clearing the `syntax-table` text property needed to prevent JSXText from introducing unterminated strings.  To solve this I will probably provide more changes (here or in another PR) that re-adds the `syntax-table` property here after it is cleared elsewhere.~~

[As noted below](https://github.com/mooz/js2-mode/pull/523#issuecomment-481984063), this now also fixes #409.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,26 @@ The stable versions are hosted at [GNU ELPA](http://elpa.gnu.org/)
 You can also install the latest development version from
 [MELPA](https://melpa.org/#/getting-started).
 
+JSX
+===
+
+Full support for highlighting and indenting of JSX is available in Emacs 27.
+Eventually it can be downloaded [here](https://www.gnu.org/software/emacs/);
+until it’s available there, you can install a snapshot of the Emacs master
+branch:
+
+```
+git clone https://git.savannah.gnu.org/git/emacs.git
+cd emacs
+./autogen.sh
+./configure
+make
+make install
+```
+
+Otherwise, in Emacs 26 and earlier, you can use `js2-jsx-mode` for rudimentary
+JSX indentation support (only).
+
 Emacs 22 and 23
 ===============
 
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -11718,15 +11718,21 @@ Selecting an error will jump it to the corresponding source-buffer error.
 (defun js2-syntax-propertize (start end)
   "Apply syntax properties from START to END."
   (goto-char start)
-  (if js-jsx-syntax (remove-text-properties start end js-jsx--text-properties))
+  (when (and
+         (boundp 'js-jsx--text-properties)
+         (bound-and-true-p js-jsx-syntax))
+    (remove-text-properties start end js-jsx--text-properties))
   (funcall
    (syntax-propertize-rules
-    ("<" (0 (ignore (if js-jsx-syntax (js-jsx--syntax-propertize-tag end))))))
+    ("<" (0 (ignore (when (and
+                           (fboundp 'js-jsx--syntax-propertize-tag)
+                           (bound-and-true-p js-jsx-syntax))
+                      (js-jsx--syntax-propertize-tag end))))))
    (point) end))
 
 ;; In Emacs >=27, this is needed for JSX highlighting.
 (defconst js2--font-lock-keywords
-  `(,@js-jsx--font-lock-keywords)
+  `(,@(bound-and-true-p js-jsx--font-lock-keywords))
   "Font lock keywords for `js2-mode'; see `font-lock-keywords'.")
 
 (defun js2-font-lock-syntactic-face-function (state)
@@ -11746,7 +11752,7 @@ Selecting an error will jump it to the corresponding source-buffer error.
   (set (make-local-variable 'indent-line-function) #'js2-indent-line)
   (set (make-local-variable 'indent-region-function) #'js2-indent-region)
   (set (make-local-variable 'syntax-propertize-function)
-       (if (version< emacs-version "27.0") nil #'js2-syntax-propertize))
+       (if (boundp 'js-jsx-syntax) #'js2-syntax-propertize nil))
   (set (make-local-variable 'comment-line-break-function) #'js2-line-break)
   (set (make-local-variable 'beginning-of-defun-function) #'js2-beginning-of-defun)
   (set (make-local-variable 'end-of-defun-function) #'js2-end-of-defun)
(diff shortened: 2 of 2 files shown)


jacksonrayhamilton:
Since I first opened the PR last night, I got some feedback via email on the code style, and made updates accordingly (a59c7b1...63b7a72).

jacksonrayhamilton:
I pushed a commit to Emacs `master` tonight (“Add extra text property to fix issue with js2-mode integration”) and I pushed a complimentary commit to this PR to fix the issue with string quotes from #409.

dgutov:
Hmm.

Have you tried using `char-property-alias-alist` instead? To create an alias for `syntax-table`, and either use it for JSX syntax properties, or in `js2-mode` everywhere instead of `syntax-table`?

jacksonrayhamilton:
Looks like that works!  Much simpler.

While I just pushed that change, I don’t mean to discount the alternative change that Stef has suggested via email.  This push is merely an improvement upon what I initially pushed.  Not sure what you think about his idea, or how to test it.

jacksonrayhamilton:
I added a section to the README which should hopefully encourage some users to download a Emacs 27 snapshot and test out my improvements.

dgutov:
After a long-ish private discussion, this is the step I have decided on for now: recommend `js-mode` together with `js2-minor-mode` as the preferred alternative. And Emacs 27, of course.

This proposal might be better in a sense (makes it easier for the users to follow), but it's harder for me to test any breakage with earlier versions of `js-mode` (especially going back to Emacs 24, which we still support), and the principle of not setting the `syntax-table` property inside the parser would introduce (probably) subtle breakage in `rjsx-mode`. So I'm going to hold off on it.

In the meantime, I'd like to know what @felipeochoa thinks about it and whether he'd like to reimplement some of the features in `rjsx-mode` in a different way, or partially merge/reuse some of the indentation code.

@jacksonrayhamilton, further comments welcome.

dgutov:
Here's hoping `js2-refactor` and friends don't rely on `js2-mode` being used exactly. They shouldn't anyway.

jacksonrayhamilton:
# Background

js-mode is now using `syntax-propertize` to set text properties that will allow for highlighting and indenting JSX code.  Since JSX is a non-standard syntax extension, we want to enable support for it cautiously, but also easily; so, js-mode is auto-detecting the presence of JSX and enabling parsing and handling of it only under certains conditions, such as:

- file name ends in “.jsx”
- if `import React from 'react'` or `var React = require('react')` appears near the beginning of a buffer

This PR would have copied some code from js-mode into js2-mode, such that js2-mode would also do those things in the same way that js-mode did them.  But, long term, there may be a better solution for js2-mode than copying code.

Currently, js2-mode uses js-mode’s code for indentation (and, in older emacsen, it backports indentation logic from Emacs 26 and earlier).  However, js2-mode doesn’t set text properties with `syntax-propertize`, nor does it highlight code with `font-lock-keywords`.  Instead, js2-mode parses the buffer and sets text properties for indenting and highlighting according to its parse.  Indentation and highlighting can be more accurate as a result, although there may also be a short delay upon initial load and after changes to the buffer before those accuracies are realized by the user.  In any case, for the time being and for the sake of consistency, it may make the most sense to keep things this way in js2-mode.

A disadvantage to the scheme described in the above paragraph is that the JSX features added in js-mode won’t be accessible when js2-mode is enabled; instead, as Dmitry has commented above, one must instead use js-mode with js2-minor-mode enabled.

# Feature Wishlist

In order for complete support for JSX to be available in js2-mode (not just js2-minor-mode), the following things should be done:

- The js2-mode parser should parse and add to its AST JSX nodes (not just E4X nodes; perhaps JSX support should even replace E4X support)
- The js2-mode parser should add text properties that the js-mode JSX indentation support (introduced in Emacs 27) can use to indent code properly
- The js2-mode parser should add font-lock text properties for highlighting JSX code properly
- js2-mode could auto-detect JSX like js-mode does

As has been pointed out by Stefan in our email thread—and I’m in agreement with him on this—long term, it may be best to set text properties related to indentation using `syntax-propertize` in js2-mode as well, rather than relying on the AST.  This would be good for performance because it would reduce the time spent after a keystroke that must elapse in order to parse the buffer sufficiently to determine how to indent nearby code (since `syntax-propertize` is chunk-based, unlike AST parsing).

Therefore, another long-term change we may wish to work towards would be:

- js2-mode should use js-mode’s `syntax-propertize-function` and not set indentation-related text properties during parsing under any conditions

However, before we make that leap, we’d like to ensure that this change will be compatible with derivative modes (like @felipeochoa’s rjsx-mode) which set `syntax-table` properties using the AST.  Of particular interest is [this line](https://github.com/felipeochoa/rjsx-mode/blob/03dd8d1683501e81b58674d64c3032b7b718402c/rjsx-mode.el#L781-L782) from rjsx-mode.

# The Future of rjsx-mode

I recognized that users were finding it challenging to work with JSX code in Emacs and also with configuring Emacs to do that.  Therefore, I resolved it would be best to include support for JSX in Emacs’ core, and that work has already been compeleted.  What does this mean for rjsx-mode?

1. Indentation and highlighting of JSX are both available to js-mode users (using Emacs 27) without needing to install rjsx-mode
2. People who use js2-mode (rather than js2-minor-mode) still need to install rjsx-mode to get indentation/highlighting
3. rjsx-mode still extends the js2-mode parser to parse JSX properly (enabling js2-mode linting of JSX code, et al)
4. rjsx-mode still provides electricity while typing tags

Given these points, I can envision the following futures for this package:

1. An effort is made to merge rjsx’s JSX parsing into js2-mode, thus point 3 (above) is obsolete; also, this creates the conditions for point 2 (above) to be resolved
2. No effort is ever made to add JSX parsing to js2-mode, and…
    1. Users gradually migrate away from JS2+RJSX and just use js-mode, or
    2. Users continue to install JS2+RJSX and we keep the status quo (which in my opinion is overly complex)

Assuming future 1 is pursued, rjsx-mode’s purpose could transition to be to provide interesting React/JSX features like electricity.  Or, electricity could also be migrated to JS2 or even to js-mode, and rjsx-mode could be retired.

But those are just my ideas, I would like to hear from @felipeochoa what he thinks of them, and how he thinks the future should unfold for his package.

DoMiNeLa10:
Relying on imports sounds like an awful idea to me, there are many libraries which rely on JSX, and looking only React (down to import names) is just a bad idea, especially because it doesn't take the possibility of a different JSX pragma (and another one for fragments) into account. These problems might crop up even in React projects that use something like `@emotion/core`, which provides its own JSX pragma.

In fact, making JS2 pragma-aware would allow to catch errors where the user forgets to import the pragma (which for React projects would be `React.createElement`).

jacksonrayhamilton:
@DoMiNeLa10 Since React is the bearer of JSX and a `React` variable usually needs to be in scope in order for `React.createElement` calls to work, I don’t think it’s a terrible starting point.

Also, Emacs 27 makes available a customizable variable called `js-jsx-regexps`.  Its default value is `("\\_<\\(?:var\\|let\\|const\\|import\\)\\_>.*?React")`.  Based on my reading of emotion’s homepage, I think you could add `"@jsx jsx"` or `"@emotion/"` to that list to improve JSX detection for your project.

jacksonrayhamilton:
Since the spirit of the changes provided by this PR won’t be merged, I’ll close it.

There are still a few changes in relation to the addition of JSX that I think should be addressed:

- Documentation within the package (rather than just the README)
- js2-jsx-mode doesn’t even indent properly anymore in Emacs 27 (nor would I expect it to, but we should figure out what to do about that)

Also, several threads could be spun from my big post above.

I’ll create separate issues / PRs for all of these things.

steinarb:
@dgutov As an emacs 26 user, It would be nice if all of the JSX PRs could be merged into js2-mode and a new release of js2-mode made to melpa.  If so, I could throw out rjsx-mode and use js2-mode instead.

dgutov:
There's no open PR that will make it all magically work in Emacs 26.

jacksonrayhamilton:
Rather, there are a number of open issues offering specifications for merging rjsx-mode and js2-mode: https://github.com/mooz/js2-mode/issues/527, https://github.com/mooz/js2-mode/issues/528, https://github.com/mooz/js2-mode/issues/529, https://github.com/mooz/js2-mode/issues/530

The rjsx-mode author is okay with his project being merged into this one.  The community simply needs a volunteer to step up and do the work.

Issue #527, opened 2019-06-04
Add JSX nodes to the JS2 AST


jacksonrayhamilton:
Continuing the discussion from [here](https://github.com/mooz/js2-mode/issues/292#issuecomment-258616453) in light of my comment from [here](https://github.com/mooz/js2-mode/pull/523#issuecomment-496288390)…

It would be nice if js2-mode added JSX nodes to its AST, like @felipeochoa’s [rjsx-mode](https://github.com/felipeochoa/rjsx-mode) does.  This issue proposes adding such a feature to js2-mode, most likely by migrating parsing code and classes from RJSX to JS2.

Such an extension would make JS2 capable of linting and highlighting JSX code.  With these capabilities available, we can justify inheriting js-mode’s capability of automatically-detecting the presence of JSX.  (These matters will be addressed in separate issues.)

Once all those features are implemented, js2-mode will behave more like a parent of js-mode.  Without them, js2-mode would be less capable of handling JSX than js-mode; js2-mode would continue to rely on rjsx-mode in order to fully support JSX.

dgutov:
At some point, adding JSX support meant doing away with supporting E4X. The latter technology seems to have very much died in the meantime, however.

PR #531, opened 2019-06-04
JSX Documentation

jacksonrayhamilton:
Since https://github.com/mooz/js2-mode/pull/523 won’t be merged, and as a follow-up to the changes you’ve made so far, I offer these additional documentation changes to compensate for the halfway-finished state of things.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # History of user-visible changes
 
-## 2019-06-03
+## Next
 
 * Emacs 27 now provides improved JSX indentation support, along with
   new JSX highlighting and detection support. Install Emacs 27 and use
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # History of user-visible changes
 
+## 2019-06-03
+
+* Emacs 27 now provides improved JSX indentation support, along with
+  new JSX highlighting and detection support. Install Emacs 27 and use
+  `js-mode` with `js2-minor-mode` (see README), rather than
+  `js2-jsx-mode`.
+* Using `js2-jsx-mode` will now trigger a warning in Emacs 27.
+
 ## 2019-02-19
 
 * Changed the default of `js2-strict-trailing-comma-warning` to nil.
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ or e.g. install a snapshot from a
 use `js-mode` as the major mode. To make use of the JS2 AST and the
 packages that integrate with it, we recommend `js2-minor-mode`. See
 the corresponding [instructions in the
-Commentary](https://github.com/mooz/js2-mode/blob/ae9fea81cf9ed80f403c161fde4d8aff1ceb3769/js2-mode.el#L57).
+Commentary](https://github.com/mooz/js2-mode/blob/d9bba5b51097e4915a36f652742588128aca3211/js2-mode.el#L57).
 
 `js-mode` in Emacs 27 includes full support for syntax highlighting
 and indenting of JSX syntax. Try it out!
--- a/README.md
+++ b/README.md
@@ -33,23 +33,10 @@ the corresponding [instructions in the
 Commentary](https://github.com/mooz/js2-mode/blob/6cc0a8a87a9178e44c51203e1fe34c57f246a7de/js2-mode.el#L60).
 
 `js-mode` in Emacs 27 includes full support for syntax highlighting
-and indenting of JSX syntax. Try it out!
+and indenting of JSX syntax.
 
-Until [#527](https://github.com/mooz/js2-mode/issues/527) is
-addressed, linting of JSX code is likely to fail. If that bothers you,
-you may elect to disable linting:
-
-```lisp
-(setq js2-mode-show-parse-errors nil)
-(setq js2-mode-show-strict-warnings nil)
-```
-
-In lieu of linting,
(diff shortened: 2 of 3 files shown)


js2-mode.el
```diff
@@ -56,16 +56,13 @@
 
 ;;   (add-hook 'js-mode-hook 'js2-minor-mode)
 
+;; If you use Emacs 27 and desire JSX support added in that release,
```
dgutov: You might want to merge this paragraph with the existing one before the code sample. Otherwise, it reads a bit weird.

Especially with the existing paragraph below which suggests to "also ... hook it", but actually refers to `js2-mode`.

README.md
```diff
@@ -30,13 +30,26 @@ or e.g. install a snapshot from a
...
+Commentary](https://github.com/mooz/js2-mode/blob/d9bba5b51097e4915a36f652742588128aca3211/js2-mode.el#L57).
 
 `js-mode` in Emacs 27 includes full support for syntax highlighting
-and indenting of JSX syntax.
+and indenting of JSX syntax. Try it out!
 
-[rjsx-mode](https://github.com/felipeochoa/rjsx-mode/) is an
-alternative option which comes with certain tradeoffs.
+Until [#527](https://github.com/mooz/js2-mode/issues/527) is
+addressed, linting is likely to fail after JSX code. If that bothers
```
dgutov: after => inside/for?

js2-mode.el
jacksonrayhamilton: Certainly, I made that change.

README.md
jacksonrayhamilton: I thought that JSX code and the code following it would not be linted correctly, but apparently the parser quickly recovers, so it’s probably only the JSX parts of the code which will show linting errors.  I tried to use clearer language in a recent commit.

js2-mode.el
```diff
@@ -51,21 +51,18 @@
...
+;; mode.  Linting of JSX code is also likely to fail.)  To install it as a minor
+;; mode:
 
 ;;   (add-hook 'js-mode-hook 'js2-minor-mode)
 
 ;; You may also want to hook it in for shell scripts running via node.js:
 
 ;;   (add-to-list 'interpreter-mode-alist '("node" . js2-mode))
 
-;; Support for JSX is available via the derived mode `js2-jsx-mode'.  If you
```
dgutov: Can we really remove this part, keeping in mind that pre-Emacs 27 it is *the* way to go?

Thinking more about it, maybe that's also the section that the Emacs 27 JSX instructions should go to instead. The "install it as a minor mode" section above gets too bloated with that extra info. Here, though, you can enumerate the cases. Use Emacs 27? Minor mode is the way to go (a note about linting problems would also be more relevant here). Use Emacs 26 or earlier? Do the stuff below.

js2-mode.el
```diff
@@ -11762,18 +11759,28 @@ Selecting an error will jump it to the corresponding source-buffer error.
...
   (add-hook \\='js2-jsx-mode-hook #\\='set-jsx-indentation)"
+  (unless (version< emacs-version "27.0")
+    ;; Emacs 27 causes a regression in this mode since JSX indentation
+    ;; begins to rely on js-mode’s `syntax-propertize-function', which
+    ;; JS2 is not currently using.
+    ;; https://github.com/mooz/js2-mode/issues/529 should address
+    ;; this.  https://github.com/mooz/js2-mode/issues/530 also has a
+    ;; piece related to the design of `js2-jsx-mode'.  Until these
+    ;; issues are addressed, ward Emacs 27 users away from this mode.
+    (display-warning 'js2-mode "For JSX support, use js-mode with js2-minor-mode"))
```
dgutov: :+1:

README.md
```diff
@@ -30,13 +30,26 @@ or e.g. install a snapshot from a
...
-Commentary](https://github.com/mooz/js2-mode/blob/ae9fea81cf9ed80f403c161fde4d8aff1ceb3769/js2-mode.el#L57).
+Commentary](https://github.com/mooz/js2-mode/blob/6cc0a8a87a9178e44c51203e1fe34c57f246a7de/js2-mode.el#L60).
 
 `js-mode` in Emacs 27 includes full support for syntax highlighting
-and indenting of JSX syntax.
+and indenting of JSX syntax. Try it out!
 
-[rjsx-mode](https://github.com/felipeochoa/rjsx-mode/) is an
-alternative option which comes with certain tradeoffs.
+Until [#527](https://github.com/mooz/js2-mode/issues/527) is
```
dgutov: I don't really mind either way, but... should this go into the wiki instead?

This section is getting a bit long, and it's not like all of our users need to read all this.

Also: if #527 is resolved, we'll probably rewrite the whole recommendation, including the use of the minor mode.

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # History of user-visible changes
 
+## 2019-06-03
```
dgutov: This should say `## Next` for now. These are dates of actual releases.

js2-mode.el
jacksonrayhamilton: Yeah.  The organization and wording needed more massaging.  I made this change.

README.md
jacksonrayhamilton: Less is more.  I’ve removed my changes here.

NEWS.md
jacksonrayhamilton: Done.

dgutov:
Terrific, thank you!

Issue #506, opened 2018-10-09
support `optional-chaining`


redguardtoo:
See https://babeljs.io/docs/en/next/babel-plugin-proposal-optional-chaining.html

PR #533, opened 2019-07-09
support optional chaining operator

redguardtoo:
See https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

and https://github.com/mooz/js2-mode/issues/506

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@
   `js-mode` with `js2-minor-mode` (see README), rather than
   `js2-jsx-mode`.
 * Using `js2-jsx-mode` will now trigger a warning in Emacs 27.
+* Support for optional-chaining operator `?.`
 
 ## 2019-02-19
 
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -635,7 +635,8 @@ which doesn't seem particularly useful, but Rhino permits it."
 (defvar js2-AWAIT 169)  ; await (pseudo keyword)
 
 (defvar js2-HOOK 170)          ; conditional (?:)
-(defvar js2-EXPON 171)
+(defvar js2-OPTIONAL-CHAINING 171) ; optional chaining (?.prop obj?.[expr] func?.())
+(defvar js2-EXPON 172)
 
 (defconst js2-num-tokens (1+ js2-EXPON))
 
@@ -1655,6 +1656,9 @@ the correct number of ARGS must be provided."
 
 (js2-msg "msg.no.colon.cond"
          "missing : in conditional expression")
+
+(js2-msg "msg.bad.optional.chaining"
+         "missing property name or [ or ( after optional chaining operator")
 
 (js2-msg "msg.no.paren.arg"
          "missing ) after argument list")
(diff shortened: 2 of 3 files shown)


js2-mode.el
```diff
@@ -10365,11 +10392,14 @@ Last token parsed must be `js2-RB'."
   (when (eq (js2-token-type token) js2-NAME)
     (js2-record-face 'js2-function-call token)))
 
-(defun js2-parse-function-call (pn)
+(defun js2-parse-function-call (pn &optional use-optional-chaining-p)
   (js2-highlight-function-call (js2-current-token))
   (js2-get-token)
   (let (args
         (pos (js2-node-pos pn)))
+    (when use-optional-chaining-p
```
dgutov: How come this is needed for optional chaining, but not for normal chaining?

js2-mode.el
```diff
@@ -6074,7 +6078,9 @@ its relevant fields and puts it into `js2-ti-tokens'."
                           (?,
                            (throw 'return js2-COMMA))
                           (??
-                           (throw 'return js2-HOOK))
+                           (if (js2-match-char ?.)
+                               (throw 'return js2-OPTIONAL-CHAINING)
```
dgutov: Looking at the parsing function, we don't seem to save any information that the chaining was optional.

It seems like we'll get exactly the same result if we simply return `js2-DOT` here and skip all other changes.

redguardtoo: @dgutov , optional chaining is new JS syntax, see https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

js2-mode.el
redguardtoo: See https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

One example is `Test?.();` instead of `Test()`

So `Test` is still a function call token but there might exist a optional chaining token between `Test` and `(params)`.

AST is like 

```
                                 function call token
                             /                |                               \
function name token   optional chaining token  paramters token
```

js2-mode.el
dgutov: I see. So maybe the patch makes sense.

But speaking of minimizing changes, you could look after `?.` and if the next char is `(`, return just that and ignore the preceding two characters. Again, if we don't store any info about optional chaining in the AST.

redguardtoo: I'm not sure I can hack the token stream. It's not the right way to do it.

Say user input `hello?.?.(` (that's wrong syntax, if I just remove `?.` after `(` from token stream. the remaining tokens becomes  `hello?.(`. So wrong usage of that operator now become syntax correct.

`use-optional-chaining-p` is not some hack. It strictly follows the standard workflow of parser,

Here is BNF of function call
FUNCTION: OPERATION  "(" [CALL_ARGS]")"
                  | OPERATION  OPTIONAL_CHAINING_OPERATOR  "(" [CALL_ARGS]")"

When `use-optional-chaining-p` is nil, pop OPERATION token only. It's the first scenario.
When `use-optional-chaining-p` is t, pop OPERATION token and one OPTIONAL_CHAINING_OPERATOR token. It corresponds to second scenario.

dgutov: OK, thanks for the explanation!

dgutov:
Could you add a NEWS entry? In NEWS.md, at the top.

redguardtoo:
@dgutov , I updated pull request.

redguardtoo:
updated

NEWS.md
```diff
@@ -8,6 +8,10 @@
   `js2-jsx-mode`.
 * Using `js2-jsx-mode` will now trigger a warning in Emacs 27.
 
+## 2019-08-14
+
+* Support for optional-chaining operator `?.`
```
dgutov: Please put this under `Next`.

redguardtoo: done

dgutov:
Thank you

PR #538, opened 2019-11-07
Allow filtering of ignored errors

garrett-hopper:
This change adds `js2-ignored-errors` which allows elements of `js2-parsed-errors` to be ignored.
My personal use case for this is to set `js2-ignored-errors` to `'("msg.bad.await")` to allow top-level `await` statements which is useful when using the `--experimental-repl-await` flag with Node REPL.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -229,6 +229,13 @@ Possible values are the keys of `js2-message-table'."
   :group 'js2-mode
   :type '(repeat string))
 
+(defcustom js2-ignored-errors nil
+  "A list of error message types that will not be reported.
+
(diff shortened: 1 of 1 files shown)


dgutov:
Any other errors can be safely ignored this way?

Unlike warnings, a presence of an error often means that the following code would be parsed correctly. So most errors couldn't be simply ignored.

Jumblemuddle:
Yeah, I'd thought about that. I don't know of any other safe errors. Even
top-level await is a little funny, since it usually is a real error.
If anyone has better ideas, I'm all ears. This just happened to solve my
problem.

Issue #537, opened 2019-11-06
Support private fields for classes with #-prefix


ArneBab:
private fields in Javascript classes (https://github.com/tc39/proposal-class-fields) are currently shown as errors.

It would be great if js2 could treat them as supported syntax.

Draft for a test:
```
(js2-deftest-parse parse-class-private-field-with-init
  "class C {\n  #x = 42;\n  #y = 24;\n}"
  :reference "class C {\n  #x = 42\n  #y = 24\n}")
```

We already use them via babel, and I get the error "invalid property id", and I don’t understand where js2-name-node-p is defined.

ArneBab:
```
(defun js2-identifier-start-p (c)
  "Is C a valid start to an ES5 Identifier?
See http://es5.github.io/#x7.6"
  (or
   (memq c '(?$ ?_ ?#))
   (memq (get-char-code-property c 'general-category)
         ;; Letters
         '(Lu Ll Lt Lm Lo Nl))))
```

ArneBab:
See https://github.com/mooz/js2-mode/pull/541

dgutov:
I'd rather not merge unfinished features, they're likely to trigger further bug reports anyway.

If you really need this for your work and you don't have the time to write a complete support, you can apply this in your own config using `advice-add`.

ArneBab:
How can I use advice-add for this? (for my own use I redefined the function)

The feature is stage 3 and there are polyfills, Chrome and Babel already support the feature: https://github.com/tc39/proposal-class-fields#implementations

Also there is support in IntelliJ and Sublime for this, but I really don’t want to use IntelliJ for Javascript, since js2-mode provides comparable JS-support in Emacs (where it is much more comfortable to work): https://github.com/tc39/proposal-class-fields/issues/57

olanod:
Any pointer on how to add support for it at least as a local hack? Looks mature enough, I'm using it in Chrome already with no transpiling, the highlighting is very ugly and confusing at the moment :(

ArneBab:
I created a pull-request that adds the support: https://github.com/mooz/js2-mode/pull/541

dgutov:
@ArneBab Could you clarify how `this.#foo = 1` is recognized instead?

dgutov:
Here's the "local hack" version:

```lisp
(advice-add #'js2-identifier-start-p
            :before-until
            (lambda (c) (eq c ?#)))
```

ArneBab:
I use this as local hack:
```
[...]

PR #541, opened 2019-11-19
Support private fields for classes with #-prefix

ArneBab:
With this change the parser no longer treats private fields in Javascript classes (https://github.com/tc39/proposal-class-fields) as errors.

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -5587,7 +5587,7 @@ Returns nil and consumes nothing if TEST is not the next character."
   "Is C a valid start to an ES5 Identifier?
 See http://es5.github.io/#x7.6"
   (or
-   (memq c '(?$ ?_))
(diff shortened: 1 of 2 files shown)


ArneBab:
This is currently missing this test:

```
(js2-deftest-parse parse-class-private-field-no-init-assigned
  "class C {\n  #x\n  \"z\"\n  456\n  _setY() {  this.#x = 1;\n}\n}"
  :warnings-count 0)
```

JeremyPouyet:
Works for me! Thanks. The possibility to customize the color would be a great addition.

ArneBab:
@JeremyPouyet please have a look here for an option with less maintenance burden: https://github.com/mooz/js2-mode/issues/537#issuecomment-578446942

PR #543, opened 2020-01-11
feat(parser): add for-await-of parsing

### LINKED ISSUE #412: Support for Async Iterators/Generators syntax
Async Iterators/Generators are at Stage 3: https://github.com/tc39/proposal-async-iteration

Great stuff, really!  js2-mode complains about `for-await-of`, and also async generators when given as methods on a class.

Fuco1:
Fixes #412

### DIFF
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -2632,14 +2632,17 @@ so many of its properties will be nil.
                                                         object
                                                         in-pos
                                                         each-pos
-                                                        foreach-p forof-p
+                                                        await-pos
+                                                        foreach-p forof-p forawait-p
                                                         lp rp)))
   "AST node for a for..in loop."
   iterator  ; [var] foo in ...
   object    ; object over which we're iterating
   in-pos    ; buffer position of 'in' keyword
   each-pos  ; buffer position of 'each' keyword, if foreach-p
+  await-pos  ; buffer position of 'await' keyword, if forawait-p
   foreach-p ; t if it's a for-each loop
+  forawait-p ; t if it's a for-await loop
   forof-p)  ; t if it's a for-of loop
 
 (js2--struct-put 'js2-for-in-node 'js2-visitor 'js2-visit-for-in-node)
@@ -2653,10 +2656,13 @@ so many of its properties will be nil.
 (defun js2-print-for-in-node (n i)
   (let ((pad (js2-make-pad i))
         (foreach (js2-for-in-node-foreach-p n))
+        (forawait (js2-for-in-node-forawait-p n))
         (forof (js2-for-in-node-forof-p n)))
     (insert pad "for ")
     (if foreach
         (insert "each "))
+    (if forawait
+        (insert "await "))
     (insert "(")
     (js2-print-ast (js2-for-in-node-iterator n) 0)
     (insert (if forof " of " " in "))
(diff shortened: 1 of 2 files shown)


js2-mode.el
```diff
@@ -2632,14 +2632,17 @@ so many of its properties will be nil.
                                                         object
                                                         in-pos
                                                         each-pos
-                                                        foreach-p forof-p
+                                                        await-pos
```
dgutov: await-pos is unused, isn't it? When printing, etc.

dgutov: Same could be said for `each-pos`, of course, but let's not worry about that.

js2-mode.el
```diff
@@ -2632,14 +2632,17 @@ so many of its properties will be nil.
...
+                                                        foreach-p forof-p forawait-p
                                                         lp rp)))
   "AST node for a for..in loop."
   iterator  ; [var] foo in ...
   object    ; object over which we're iterating
   in-pos    ; buffer position of 'in' keyword
   each-pos  ; buffer position of 'each' keyword, if foreach-p
+  await-pos  ; buffer position of 'await' keyword, if forawait-p
   foreach-p ; t if it's a for-each loop
```
dgutov: We could also set `foreach-p` to a symbol (e.g. `AWAIT`), instead of adding a new field. But I'm okay with either.

dgutov:
Could you add a NEWS entry as well?

js2-mode.el
Fuco1: I can remove them both, I think they aren't necessary.  But maybe some 3rd party depends on it, who knows.

js2-mode.el
Fuco1: I did it this way to not possibly break existing integrations outside this package.  If we're okay with removing backward compatibility we can use as symbol as you say.

Maybe in the future they will allow `for each await ()` ? :D

Fuco1:
@dgutov will do thanks for the feedback

js2-mode.el
dgutov: Up to you.

js2-mode.el
dgutov: Integrations that use an `(eq ... t)` check? I wouldn't be too worried about those.

Would `for each await` actually make any kind of sense?

Fuco1: Rather even `(if foreach-p ...)` which would now be set to a symbol.  But of course we would rename the field so I guess in fact that mitigates the problem and should allow an easy migration path.

I'm honestly not even sure what `for each` does.  So probably not, never used it in my life.

dgutov: Actually if we don't rename the field, it should make the "migration" easier: only dependents that know about "for await" would recognize it, and the rest would continue to compile and treat the AST nodes as "for each". Which is the outdated syntax.

It was Mozilla's extension. The modern counterpart is "for of": https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Statements/for_each...in

dgutov: Again, I'm fine with either (rename of not). But reusing the field seems more economical.

pcr910303:
@Fuco1 May I ask the status of this? I would very much like if this gets merged...

Fuco1:
@pcr910303 The status is that I forgot about it :blush: 

@dgutov is the news entry the only thing blocking this?

dgutov:
@Fuco1 Pretty much.

The other two discussions don't seem critical.

wyuenho:
Jesus, please just merge this and add the NEWS entry in a separate commit. There's absolutely no need to be bureaucratic here.

dgutov:
Okay.

dgutov:
I don't know if it was that unreasonable to expect the contributor to do at least one revision after review, but oh well.

wyuenho:
It is unreasonable if you hold up perfectly good progress for over a year due to a missing NEWS entry that nobody reads and you can add yourself.

dgutov:
Next time you have an idea about what somebody else should do, try helping.

wyuenho:
Help how? You are the maintainer, not me and I don't care about the NEWS entry. I commented, you got a notification, the code got merged after over a year. That's helping everyone.

dgutov:
> You are the maintainer, not me and I don't care about the NEWS entry.

So rather than pitch in to maintain the project standards, you chose to be rude instead.

Thanks for the reminder, but the rest could have been done better.

wyuenho:
I'll make sure I point out the obviously unreasonableness of holding up a perfectly good PR for over year next time. Without any exclamations involving any deities. I promise.

PR #544, opened 2020-02-11
Use type face for class

wyuenho:
Because now we can, and it's consistent with how every other major modes for languages that have classes.


dgutov:
Sounds good.

PR #550, opened 2020-05-22
Fix ‘huge-hex’ test for newer Emacsen.

phst:
Starting with Emacs 27, we have big integers, so the huge hexadecimal value is
parsed exactly.


dgutov:
Thanks!

PR #551, opened 2020-05-22
Avoid undefined behavior in indentation tests.

phst:
Info node ‘(elisp) Local Variables’ states:

> […] the order of _bindings_ is unspecified: […] Therefore, avoid binding a
> variable more than once in a single ‘let’ form.

In fact, the tests that bind ‘js2-pretty-multiline-declarations’ twice fail if
indent.el is byte-compiles.


dgutov:
What if you just replace `let` with `let*`?

phst:
> What if you just replace `let` with `let*`?

Yes, that also works.

dgutov:
Thanks!

PR #576, opened 2021-10-31
Support mocha-like test files in Imenu

### LINKED ISSUE #575: Imenu support for mocha-like test files
I've created [js2-imenu-mocha](https://github.com/DamienCassou/js2-imenu-mocha) which adds `describe()` and `it()` blocks to imenu entries. Would it make sense to include that in `js2-mode`'s repository directly? Maybe as part of `js2-imenu-extras`?

DamienCassou:
Closes #575.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+* Imenu support for mocha-like (includes Jasmine and Cypress) test
+  files: i.e., `M-x imenu` will now list test blocks defined with
+  `describe()` and `it()`
+  ([#576](https://github.com/mooz/js2-mode/pull/576)).
 * Minor improvements in `js2-jump-to-definition`
   ([#423](https://github.com/mooz/js2-mode/issues/423)).
 * Support for private class members
--- a/js2-imenu-extras.el
+++ b/js2-imenu-extras.el
@@ -35,6 +35,10 @@
 (require 'cl-lib)
 (require 'js2-mode)
 
+(eval-when-compile
+  (when (<= emacs-major-version 26)
+    (require 'subr-x)))
+
 (defvar js2-imenu-extension-styles
   `((:framework jquery
      :call-re   "\\_<\\(?:jQuery\\|\\$\\|_\\)\\.extend\\s-*("
@@ -59,6 +63,14 @@
     (:framework react
      :call-re "\\_<React\\.createClass\\s-*("
      :recorder js2-imenu-record-react-class)
+
+    (:framework mocha
+     :call-re ,(rx line-start
+                   (* (syntax whitespace))
+                   (or "describe" "fdescribe" "describe.only")
+                   (* (syntax whitespace))
+                   "(")
+     :recorder js2-imenu-record-mocha-describe)
 
     (:framework sencha
      :call-re "^\\s-*Ext\\.define\\s-*("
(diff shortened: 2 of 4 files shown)


js2-imenu-extras.el
```diff
@@ -221,6 +229,161 @@ Currently used for jQuery widgets, Dojo and Enyo declarations."
                                      (list name-value))
                                    (js2-node-abs-pos methods))))))
 
+(defun js2-imenu-record-mocha-describe ()
+  "Populate `js2-imenu-recorder' with mocha-like describe/it/beforeEach/… nodes."
+  (let ((node (js2-node-at-point (1- (point)))))
+    (when (js2-imenu-extras--mocha-top-level-describe-p node)
+      (js2-imenu-extras--mocha-visit-node node (list)))))
```
dgutov, line 236: FWIW, `(list)` evaluates to `nil`.

dgutov:
Solid effort :+1:. Could you rebase it on master now?

js2-imenu-extras.el
```diff
@@ -221,6 +229,161 @@ Currently used for jQuery widgets, Dojo and Enyo declarations."
...
+  (js2-visit-ast
+   node
+   (lambda (child end-p)
+     (when (not end-p)
+       (js2-imenu-extras--mocha-check-unknown-node child qname)))))
+
+(defun js2-imenu-extras--mocha-check-unknown-node (node qname)
+  "If NODE is a mocha test block, populate `js2-imenu-recorder'.
+
+QNAME is a list of nodes representing the qualified name of
```
dgutov, line 257: The description of QNAME is repeated multiple times in other functions, the whole paragraph. Appreciate the good documentation, but it looks a little repetitive.

How about we keep it in `js2-imenu-extras--mocha-visit-node`, but in other functions say

    QNAME is the same as described in `js2-imenu-extras--mocha-visit-node'.

?

dgutov:
A NEWS entry would be nice to have, too.

js2-imenu-extras.el
DamienCassou: IMO, `(list)` better conveys the meaning of the parameter being a list than `nil`. I'm fine changing it to `nil` if you want me to.

DamienCassou:
I took care of your feedback but there are issues with the tests in 25.1. Do you have any idea?

js2-imenu-extras.el
dgutov: It's all right.

dgutov:
> but there are issues with the tests in 25.1. Do you have any idea?

Looks like `when-let*` was not in Emacs 25. I think if you just change it to `when-let`, everything will work.

NEWS.md
```diff
@@ -11,6 +11,10 @@
 * Support for trailing comma in arrow function parameters
   ([#480](https://github.com/mooz/js2-mode/issues/480)).
 * Support for `for await of` syntax ([#543](https://github.com/mooz/js2-mode/pull/543)).
+* Imenu support for mocha-like (includes Jasmine and Cypress) test
```
dgutov, line 14: Please move the entry to the top: they all go in reverse-chronological order here.

DamienCassou: fixed

DamienCassou:
> Looks like when-let* was not in Emacs 25. I think if you just change it to when-let, everything will work.

Indeed. I also had to `require` "subr-x". I also added Emacs 26.3 to GitHub Actions because it also needs "subr-x".

js2-imenu-extras.el
```diff
@@ -35,6 +35,10 @@
 (require 'cl-lib)
 (require 'js2-mode)
 
+(eval-when-compile
+  (when (<= emacs-major-version 26)
+    (require 'subr-x)))
```
dgutov, line 40: The version check wasn't really necessary, FWIW.

`when-let` is autoloaded in newer Emacs, but it doesn't hurt to require the package too.

dgutov:
Thanks!

DamienCassou:
Thank you for your reviews and the package!

PR #577, opened 2021-11-01
Add GitHub Actions workflow

DamienCassou:
Unit tests will now be executed for every PR.


dgutov:
Thanks, Damien!

The current Travis config has indeed been broken for some time. But it's pretty fun to realize that when it worked, it was using a PPA of yours :thought_balloon:

PR #587, opened 2022-04-29
Use Eask for CI?

jcs090218:
This patch does the following:

#### In `.github/workflow/emacs.yml`

* Use `jcs090218/setup-emacs` instead of `purcell/setup-emacs`
  - Add retry capability to prevent failure on Emacs' installation
  - Support windows
* Add setup-node (`16`)
* Add `setup-eask`
* Add test for Emacs `28.1`, drop Emacs `25.1` (Eask must use `26.1` or above)
* Add tests on `Windows` and `macOS`

#### In `Makefile` (newly added)

* Replace Emacs command-line arguments `emacs --batch ...` with just `eask <command>`

> Edit:

* Add `make checkdoc`
* Add `make lint` (but left unused for now)

#### Others

* Add `Eask`-file

---

This is totally optional, I think the solution may not be ideal since this drops the test on Emacs `25.1`. :(

Let me know what you think? Hope this could make CI on Emacs a lot easier! :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
 
     - uses: actions/setup-node@v2
       with:
-        node-version: '14'
+        node-version: '16'
 
     - uses: emacs-eask/setup-eask@master
       with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   test:
-    if: github.repository_owner == 'mooz'
+    #if: github.repository_owner == 'mooz'
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,19 +15,30 @@ on:
 jobs:
   test:
     if: github.repository_owner == 'mooz'
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


dgutov:
Hi!

Thanks for the effort, but could you explain what exactly this will make easier? `js2-mode` has no package dependencies, and the test script is easily invoked through `make`. And the diff adds more lines than it deletes.

So what are we getting here?

jcs090218:
Hi, thanks for your time to review this PR!

Eask shares the same idea with Cask. I guess their manual page explained better than I do, see [Why Cask?](https://cask.readthedocs.io/en/latest/guide/introduction.html#package-development). Furthermore from this PR, Eask tries to eliminate the test command like `emacs --batch -Q ...` to **more** human readable command `eask compile`, or `eask ert`, etc. 

And you are correct, it indeed adds more lines than it deletes. I think the biggest addition lines are came from `Eask` (+14) and `setup-node` and `setup-eask` from the `test.yml` (+8). I have also added `checkdoc` and `lint` in the Makefile (sorry, I should've mentioned in the first post). With Eask, you can easily expand/eliminate your needs.

Hope these answered your questions!

To reject this PR is understandable! Since I think the Cask philosophy isn't for everyone, everything should be good! :)

dgutov:
I guess one thing we have from the "no Cask" list is no dependencies. And the existing `Makefile` rules, combined with the CI, make it fairly easy for the contributors, I think. But the argument for consistency is not bad. As long as the ecosystem standardizes on this, I guess.

I'll take a look at the added rules and play with Eask when I have the time. In the meantime, let's keep this open, if you don't mind.

Thanks!

PR #589, opened 2022-06-17
Remove the correct hook in js2-minor-mode-exit

pkryger:
The `js2-minor-mode-enter` adds `js2-minor-mode-edit` to `after-change-functions`. Contrary the `js2-minor-mode-exit` removes `js2-mode-edit` which had not been added to the list.

Without that, after enabling `js2-minor-mode` the `js-lint` remains enabled even when the mode exits.

I decided to go straight for a PR, as the issue looks like an obvious mistake, at first glance. Should that be not the case, apologies, for a naïve approach.


dgutov:
Thanks!

PR #590, opened 2022-06-25
Add support for the d (hasIndices) and s (dotAll) regexp flags

DamienCassou:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Next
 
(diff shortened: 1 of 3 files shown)


dgutov:
Thanks! Could you add a NEWS entry too?

DamienCassou:
> Thanks! Could you add a NEWS entry too?

done

PR #593, opened 2022-10-28
Support top-level await

DamienCassou:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await

This is part of ECMAScript 2022: https://en.wikipedia.org/wiki/ECMAScript#ES2022

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Next
 
+* Support top-level await
+  ([#593](https://github.com/mooz/js2-mode/issues/593)).
 * Fix for escaping inside template literal
   ([#592](https://github.com/mooz/js2-mode/issues/592)).
 
--- a/js2-mode.el
+++ b/js2-mode.el
@@ -8064,7 +8064,7 @@ string is NAME.  Returns nil and keeps current token otherwise."
                nil)
            ;; The parse was successful, so process and return the "await".
            (js2-record-face 'font-lock-keyword-face current-token)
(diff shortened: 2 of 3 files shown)


dgutov:
Thanks!

jordigh:
Hm, this only works inside ECMAscript modules, though. It should still be an error inside CommonJS files. I can't see anything in js2-mode about distinguishing the two kinds of javascript files, though.

DamienCassou:
@jordigh My PR was only about the syntactical aspect of the problem. We could probably introduce some semantic check as well as JS2 as some already. My opinion is that semantic validation
shouldn't be the responsibility of a major mode but that's just my opinion. If you care you might want to open an issue.

jordigh:
I mean, it's a syntax error if the file isn't an ECMAscript module. It would be like saying that `new` is a valid keyword in a C file because you don't have a mechanism to tell C and C++ files apart.

dgutov:
Is there a way to distinguish at the lexical or the major mode level?

jordigh:
No, unfortunately, in order to be able to tell if a file is an ESM file or a CJS file you have to look at either the filename (.cjs is guaranteed CJS and .mjs is guaranteed ESM), or you have to parse the nearest `package.json` to see what is the correct file type.

dgutov:
Someone motivated could add that check (which would look at the extension, and then also look around for `package.json` and its contents). As long as that won't bring false negatives in the end.

PR #600, opened 2024-03-10
Cleanup some deprecated code

DamienCassou:
This only replaces some deprecated code as js2-mode still declares
Emacs 24.1 as being supported.


dgutov:
Thanks!

PR #604, opened 2024-09-08
fix parsing export { foo as "bar" }

mishoo:
Fixes parsing `export { foo as "some name" }`.

https://github.com/mishoo/UglifyJS/issues/5935#issuecomment-2335157622


dgutov:
Thanks!

PR #605, opened 2024-09-09
A bunch of fixes for `export` nodes

mishoo:
`js2-visit-extern-binding` will now walk the bindings in proper order (left-to-right) both for `import` and `export` parents. Previously, the walker would descend into the right node first for `export`, with the visible bug that `js2-node-at-point` would not find the `foo` name in this case:

    export { |foo as bar };

Other small fixes:

- do not extend the boundary of the `foo` name node in a case like this:

      export default foo     ;

- include the semicolon when printing nodes like the above

- fix a bunch of warnings related to length of docstrings; also type of a defcustom.

- minor cleanups


dgutov:
Hi! Thanks for this.

Could you add a few more tests? The `js2-deftest-parse` macro should at least help exercise the change in the printing behavior.

mishoo:
> Could you add a few more tests? The `js2-deftest-parse` macro should at least help exercise the change in the printing behavior.

Yep, that's why I marked it as draft, but I didn't get to it yet... hopefully this week.

